PR #29, opened 2022-07-12
added initial hazardlink Object

gowerc:
Dependent on #30


R/HazardLink.R
```diff
@@ -0,0 +1,76 @@
...
+    function(x, y) standardGeneric("merge")
+)
+
+
+setMethod(
+    f = "merge",
+    signature = "HazardLink",
+    definition = function(x, y) {
+        HazardLink(
+            stan = merge(x@stan, y@stan),
```
G-Kazantzidis, line 70: This implies that we are going to create a second method for `merge` for the `StanModule` object ?

G-Kazantzidis (approved):
Thank you for creating the HazardLink object and its creators

gowerc:
@G-Kazantzidis , sorry just to say this is now ready for review

R/hazard_link.R
```diff
@@ -0,0 +1,79 @@
...
+)
+
+
+#' @importFrom assertthat assert_that
+#' @rdname StanModule-class
+#' @export
+setMethod(
+    f = "initialize",
+    signature = "HazardLink",
+    definition = function(.Object, ..., stan = StanModule(), contribution, parameters) {
```
G-Kazantzidis, line 37: `stan = StanModule()` should we further specify this to produce a default `HazardLink` ?

R/hazard_link.R
```diff
@@ -0,0 +1,79 @@
...
+            length(parameters) >= 1,
+            msg = "`parameter` must be a character vector"
+        )
+        
+        assert_that(
+            is.character(contribution),
+            length(contribution) == 1,
+            msg = "`contribution` must be length 1 character vectors"
+        )
+
```
G-Kazantzidis, line 50: Suggestion: 
`assert_that(names(stan@priors) == parameters)`
Moreover, we can create a function to define a `normal(0,1)` prior for the link parameter if the user does not define the prior.

gowerc: I think we discussed this in yesterdays call but my feeling is that this might be too restrictive. For example it may be that users don't want to define priors and let stans default prior be used or perhaps they want to define priors for secondary parameters required for functions which aren't used in the hazard link (though this is likely bad practice).

R/hazard_link.R
gowerc: What would the default be?

My thinking is that this is more of a fundamental object line `StanModule()` itself but doesn't really have a natural default outside of its model specific implementations i.e.  `ExponentialHazardLink()`

PR #30, opened 2022-07-13
Add stan module

G-Kazantzidis:


R/testfile.R
```diff
@@ -0,0 +1 @@
+this <- c("hi")
```
gowerc, line 1: Not sure why this has been added :)

R/stan_module.R
```diff
@@ -0,0 +1,86 @@
...
+#' @importFrom assertthat assert_that
+#' @param functions TODO
+#' @param data TODO
+#' @param parameters TODO
+#' @param transformed_parameters TODO
+#' @param prior TODO
+#' @param generated_quantities TODO
+#' @param inits TODO
+#' @export
+stan_module <- function(functions, data, parameters, transformed_parameters,
```
gowerc, line 49: Instead of an explicitly function I think instead we should be using the "initialise" method. This is a magic method that is automatically called when users run the `StanModules()` function

R/stan_module.R
```diff
@@ -0,0 +1,86 @@
...
+        inits = "list"
+    )
+)
+
+
+
+#' source_stan_part exports a .stan file as a character
+#' @param filename The name of the .stan file in the directory of the package.
+#' @export
+source_stan_part <- function(filename) {
```
gowerc, line 30: I think there needs to be logic here that can handle the scenario of the user actually providing the code directory instead of a file path something like (pseudo code):

```
function (string) {
    if  is_file(string):
        return  read_file(absolute_filename)
    else:
        return string
}
```

G-Kazantzidis: I use `file.exists` to check whether the string is an existing file and otherwise I use the `system.file` function to locate the file.

R/stan_module.R
```diff
@@ -0,0 +1,93 @@
...
+        absolute_filename <- filename
+    } else {
+        absolute_filename <- system.file(
+            "stanparts",
+            filename,
+            package = "jmpost"
+        )
+    }
+
+    readLines(absolute_filename)
```
gowerc, line 41: Sorry this logic doesn't make sense to me.   If the user were to provide the stan code as a string i.e.

```
parameters = "
    real beta;
    real gamma;
    real sigma;
"
```

This would error because line 34 would try and find it as as a system file ?

R/stan_module.R
```diff
@@ -0,0 +1,93 @@
...
+#' @param prior TODO
+#' @param generated_quantities TODO
+#' @param inits TODO
+#' @export
+setMethod(
+    f = "initialize",
+    signature = "StanModule",
+    definition = function(.Object, ..., functions, data,
+                          parameters, transformed_parameters,
+                          prior, generated_quantities, inits) {
```
gowerc, line 59: Should this be `priors` to be consistent with the pluralisation we've put on the other function argument names ?

R/stan_module.R
```diff
@@ -0,0 +1,93 @@
...
+            is.character(functions),
+            length(functions) == 1,
+            is.character(data),
+            length(data) == 1,
+            is.character(parameters),
+            length(parameters) == 1,
+            is.character(transformed_parameters),
+            length(transformed_parameters) == 1,
+            is.character(gnenerated_quantities),
+            length(gnenerated_quantities) == 1,
```
gowerc, line 70: I think its fine for this to be length 0, maybe initialise the default arguments with just `""`.  Because I can imagine it being valid to create a stan block which doesn't have one of the sections filled in, for example the HazardLink objects aren't going to need to populate the data block or the generated_quatities block ?

R/stan_module.R
```diff
@@ -1,44 +1,55 @@
...
+  "StanModule",
+  representation(
+    functions = "character",
+    data = "character",
+    parameters = "character",
+    transformed_parameters = "character",
+    priors = "list",
+    generated_quantities = "character",
+    inits = "list"
+  )
```
gowerc, line 22: You seem to have reverted to only 2 space indents here ?

R/stan_module.R
G-Kazantzidis: Indeed, allowing the user to add the code as a text could be helpful. Although I will keep the option of the `system.file` as well. It is important because we want to locate the examples we are going to provide along with the package.

R/stan_module.R
```diff
@@ -1,44 +1,55 @@
...
-#' @param filename The name of the .stan file in the directory of the package.
+#' read_stan_part returns stan code as a character.
+#' @param file Character, either the absolute path of a stan file, or the name of the stan file in the package directory or the stan code as a string.
 #' @export
-source_stan_part <- function(filename) {
-    if (file.exists(filename)) {
-        absolute_filename <- filename
+read_stan_part <- function(file) {
+    if (file.exists(file)) {
+        absolute_filename <- file
```
gowerc, line 32: I think this line can be deleted and just pass `file` to the `readLines` function

R/stan_module.R
```diff
@@ -1,44 +1,55 @@
...
+        out <- readLines(absolute_filename)
+
     } else {
-        absolute_filename <- system.file(
-            "stanparts",
-            filename,
-            package = "jmpost"
-        )
+
+        if (file.exists(system.file("stanparts", file, package = "jmpost"))) {
```
gowerc, line 37: Perhaps collapse into a single ` } else if {` block. At least I'm not sure the nested if is needed ?

R/stan_module.R
```diff
@@ -1,44 +1,55 @@
...
+                "stanparts",
+                file,
+                package = "jmpost"
+            )
+
+            out <- readLines(absolute_filename)
+
+        } else {
+
+            out <- filename
```
gowerc, line 48: Does this not error because the argument is now just `file` ? Perhaps rename as `string` or something generic to avoid confusion as it can now be either a string or a filename

R/stan_module.R
```diff
@@ -47,47 +58,42 @@ source_stan_part <- function(filename) {
...
+                        parameters, transformed_parameters,
+                        priors, generated_quantities, inits) {
+    assert_that(
+      is.character(functions),
+      is.character(data),
+      is.character(parameters),
+      is.character(transformed_parameters),
+      is.character(generated_quantities),
+      msg = "`Functions`, `data`, `parameters`, `transformed_parameters` and `gnenerated_quantities` must be character vectors"
+    )
```
gowerc: Looks like you have a mix of 4 and 2 spacing here

R/stan_module.R
```diff
@@ -47,47 +58,42 @@ source_stan_part <- function(filename) {
...
-            is.list(prior),
-            is.list(inits),
-            length(inits) == length(prior),
-            msg = "`Prior` and `inits` must be list of the same length"
-        )
+    assert_that(
+      is.list(priors),
+      is.list(inits),
+      length(inits) == length(priors),
+      msg = "`Priors` and `inits` must be list of the same length"
```
gowerc: mix of 4 and 2 spacing

R/stan_module.R
```diff
@@ -47,47 +58,42 @@ source_stan_part <- function(filename) {
...
+      ...,
+      functions = read_stan_part(functions),
+      data = read_stan_part(data),
+      parameters = read_stan_part(parameters),
+      transformed_parameters = read_stan_part(transformed_parameters),
+      priors = priors,
+      generated_quantities = read_stan_part(generated_quantities),
+      inits = inits
+    )
+  }
```
gowerc: Looks like this is just 2 spacing

R/stan_module.R
```diff
@@ -0,0 +1,99 @@
...
+#' @param inits TODO
+#' @export
+setMethod(
+    f = "initialize",
+    signature = "StanModule",
+    definition = function(.Object, ..., functions, data,
+                          parameters, transformed_parameters,
+                          priors, generated_quantities, inits) {
+        assert_that(
+            is.character(functions),
```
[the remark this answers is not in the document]
G-Kazantzidis: I added the `paste_str` function to paste the elements of a vector if the user provides a vector

R/stan_module.R
```diff
@@ -0,0 +1,126 @@
...
+read_stan_part <- function(file) {
+    if (file.exists(file)) {
+        absolute_filename <- file
+        out <- readLines(absolute_filename)
+    } else if (file.exists(system.file("stanparts", file, package = "jmpost"))) {
+        absolute_filename <- system.file(
+            "stanparts",
+            file,
+            package = "jmpost"
+        )
```
gowerc, line 39: You end up calling the system.file function twice here with identical arguments, perhaps just call it once in advance and assign the value to a variable.

R/stan_module.R
```diff
@@ -0,0 +1,126 @@
...
+            inits = inits
+        )
+    }
+)
+
+
+#' paste_vector Returns a single string from a vector of strings
+#' @param vec A vector of stings representing stan code.
+#' @export
+paste_str <- function(vec) {
```
gowerc, line 116: It would be good to add some unit tests to show this function is working as expected

R/stan_module.R
```diff
@@ -0,0 +1,126 @@
...
+
+    # check if all elements of vector include ";"
+    if (all(grepl(";", vec, fixed = TRUE) == TRUE)) {
```
gowerc, line 119: Shouldn't this be a check that the line ends with ";"  for example what about the string `real sigma; real beta`

gowerc: Also the `== TRUE` is not needed

R/stan_module.R
```diff
@@ -0,0 +1,126 @@
...
+        paste0(paste0(vec, collapse = "\\n "), "\\n")
+    } else if (all(grepl(";", vec, fixed = TRUE) == FALSE)) {
+        paste0(paste0(vec, collapse = ";\\n "), ";\\n")
+    } else {
```
gowerc, line 123: I'm not sure its even possible to enter the `else` block here is it ?

R/stan_module.R
G-Kazantzidis: > Also the `== TRUE` is not needed

but then, if `grepl(";", vec, fixed = TRUE)` gives only `FALSE`, the if condition will be `TRUE`.

R/stan_module.R
gowerc: ah I see when we have differences in some having semi colons and others not gotcha.

R/stan_module.R
```diff
@@ -0,0 +1,126 @@
...
+        assert_that(
+            is.list(priors),
+            is.list(inits),
+            length(inits) == length(priors),
+            msg = "`Priors` and `inits` must be list of the same length"
+        )
+
+        if (length(functions) > 1) {
+            functions <- paste_str(functions)
+        }
```
gowerc, line 83: Subjective but I'm wondering if it makes sense to move the  `if(length(...)> 1` logic inside of the `paste_str` function so that we don't have to have these repeated `if()` blocks ?

R/stan_module.R
```diff
@@ -0,0 +1,126 @@
...
+            is.character(parameters),
+            is.character(transformed_parameters),
+            is.character(generated_quantities),
+            msg = "`Functions`, `data`, `parameters`, `transformed_parameters` and `generated_quantities` must be character vectors"
+        )
+
+        assert_that(
+            is.list(priors),
+            is.list(inits),
+            length(inits) == length(priors),
```
gowerc, line 77: I think we need to allow for `priors` being populated and `inits` being left blank. This currently forces you to define your own `init` even if you'd rather leave it up to Stan to do that for you

R/stan_module.R
```diff
@@ -0,0 +1,126 @@
...
+#' @param data TODO
+#' @param parameters TODO
+#' @param transformed_parameters TODO
+#' @param priors TODO
+#' @param generated_quantities TODO
+#' @param inits TODO
+#' @export
+setMethod(
+    f = "initialize",
+    signature = "StanModule",
```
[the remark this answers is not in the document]
G-Kazantzidis: Okay, I am not an S4 expert either. I agree with this approach.

R/stan_module.R
G-Kazantzidis: Another option would be to allow for `length` to be `>1`. And we can paste it together later.

R/stan_module.R
G-Kazantzidis: The idea is that the default `inits` will be a list of either `NAs`, values or `functions`. This list will be curried on until the moment when the data are merged to the same object as the code. Then, the functions within the `inits` list will generate the necessary initial values.  For example the default list of the initial values of the os model https://github.com/Genentech/jmpost/blob/c9b971708856990559d88e31026a2402b1bf0fd9/R/inits.R#L25-L29

R/stan_module.R
gowerc: Potentially yes we could handle this in the is.character method.  Would also make merging StanModules easier as we don't need to worry about the logic of collapsing them there.

R/stan_module.R
gowerc: I think it would be good to assign sensible defaults here  to avoid users running into this error which might be quite annoying, perhaps something like:

```
    definition = function(
        .Object,
        <other arguments>
        priors = list(),
        inits =  rep(NA, length(priors))
    )
```

G-Kazantzidis: >         inits =  rep(NA, length(priors))

The issue here is that the inits list might contain an element that is a list or a vector of length defined from the data object. Thus, the stan code will expect eg `inits$beta_os_cov` to be a vector of length 5 and it will get only one `NA`.

gowerc: Fair point, maybe instead we should just support `inits` being a single `NA` value which we can then replicate across later when we reduce the code

G-Kazantzidis: What about being a list of either `NA` values or functions? I do believe that there is a benefit of carrying this list of values or functions from the beginning. This way, the user can modify this list at any point.

gowerc: O yes, sorry I'm not saying not support functions / list of values, what I'm trying to implement is a simple way for the user to say "I don't want to set priors for anything"  which I think just providing a single NA value (or a null or a blank list) would be quite clean.

Thinking about it I think a blank list is fine,  I  am thinking that if any parameter is missing from inits then we should assume they dont want to specify a starting value for that parameter (assuming stan can support this)

gowerc:
I'm currently getting an error when I try to use the object using non-file code btw: 

```
> StanModule(
+     functions = "myfun(){awdawda}"
+ )
Error in file.exists(file) : invalid 'file' argument
```

R/stan_module.R
```diff
@@ -157,3 +157,29 @@ setMethod(
...
+            functions = paste0(x@functions, "\\n ", y@functions, "\\n "),
+            data = paste0(x@data, "\\n ", y@data, "\\n "),
+            parameters = paste0(x@parameters, "\\n ", y@parameters, "\\n "),
+            transformed_parameters = paste0(x@transformed_parameters, "\\n ", y@transformed_parameters, "\\n "),
+            priors = append(x@priors, y@priors),
+            generated_quantities = paste0(x@generated_quantities, "\\n ", y@generated_quantities, "\\n "),
+            inits = append(x@inits, y@inits)
+        )
+    }
+)
```
gowerc, line 185: @G-Kazantzidis , sorry I thought we said in another chat that we would handle concatination in the as.character method and that we would suppoer  data/functions/etc  being > length 1.  In this case I would have expected this to just be a combining of vectors  i.e.  

```
functions =  c(x@functions, y@functions)
```

and then we have  a single place to handle  `\\n` and `;` in the as.character method?

R/stan_module.R
```diff
@@ -157,3 +157,29 @@ setMethod(
...
+#' @param x First StanModule
+#' @param y Second StanModule
+#' @export
+setGeneric(
+    "merge",
+    function(x, y) standardGeneric("merge")
+)
+
+
+setMethod(
```
gowerc, line 171: I think you need to export this method

R/stan_module.R
G-Kazantzidis: I see your point. Lets allow the user to provide an empty list at this stage.

R/stan_module.R
```diff
@@ -76,6 +77,11 @@ setMethod(
             msg = "`Priors` and `inits` must be lists"
         )
 
+        validate_that(
+            length(inits) == length(priors),
+            msg = "Lengths of inits and priors differ"
+        )
+
```
gowerc, line 84: How come you've used `validate_that` instead of `assert_that`, don't we want to throw an error if the condition isn't met ?

G-Kazantzidis: Because of the case that the user defines the prior list (which is essential for the model built) but not the initial values. In that case, the lengths of the two lists are going to be different.

gowerc (approved):

Issue #39, opened 2022-07-26
`read_stan` function does not accept empty charachter strings


G-Kazantzidis:
Bug: The `read_stan` function returns an error if the input of the function is a empty character string (`""`).
Reason: Command `system_file <- system.file("stanparts",  "', package = "jmpost") ` as in line 32 below ( where `string = ""`), returns the full path to the stanparts folder in the `jmpost` local repository. Although this path is not a file, when it is given as an input to the function `file.exists()`, the output is `TRUE` (since the path exists). Thus, in the flow of the function as the link below, the first `if` condition is rejected (line 33) and the second one is accepted (line 35) (whereas the correct one in this case is the third - line 37). 

https://github.com/Genentech/jmpost/blob/c65a61c06c4c22c85e0b8490fb52e506373fba41/R/stan_module.R#L31-L41

I fix this bug by replacing the `file.exists` in line 35 with `is_file` from the package `fs`.

gowerc:
rather than depending on another package I think we could handle this by also checking that its not a directory:

```
> file.exists("local/testing.stan")
[1] TRUE
> dir.exists("local/testing.stan")
[1] FALSE
>
```

G-Kazantzidis:
> is_file <- function(filepath) {
>     return( file.exists(filepath) & !dir.exists(filepath))
> }

I like this solution better. Lets proceed with this one.

PR #41, opened 2022-07-27
fixed multiple line bug (#37)

### LINKED ISSUE #37: StanModule errors if file contains more than 1 line
Not tested this yet but based on a code review we believe the StanModule object will error if it tries to load a Stan file that contains more than 1 line of code.

This error should be fixed and a unit test put in place to show that it is fixed.

gowerc:
Closes #37


R/stan_module.R
```diff
@@ -31,9 +32,9 @@ StanModule <- setClass(
 read_stan <- function(string) {
     system_file <- system.file("stanparts", string, package = "jmpost")
     if (file.exists(string)) {
-        out <- readLines(string)
+        out <- read_file(string)
     } else if (file.exists(system_file)) {
```
G-Kazantzidis, line 36: @gowerc is it possible to replace the `file.exists(system_file)` with `is_file(system_file)`. `is_file` is a function from package `fs` and explicitly checks whether a file exists and not whether the provided path is TRUE. I created a relevant issue [here ](https://github.com/Genentech/jmpost/issues/39).

gowerc:
@G-Kazantzidis I added the code here to resolve #39  as well apologies I got confused on wether you were doing it or if you were asking me to address it but it felt related so I threw it in here

R/stan_module.R
```diff
@@ -23,17 +23,18 @@ StanModule <- setClass(
 )
 
 
+
 #' read_stan returns stan code as a character.
 #'
 #' @param file Character, either the absolute path of a stan file, or the name of the stan
 #' file in the package directory or the stan code as a string.
```
G-Kazantzidis, line 30: #' @importFrom fs is_file

R/stan_module.R
```diff
@@ -212,3 +213,37 @@ remove_blank_strings <- function(x) {
...
+
+#' Read entire file as a single string
+#'
+#' Simple utility function to read in a file as 1
+#' continous string
+#'
+#' @param filename Location of file to read in
+#'
+read_file <- function(filename) {
+    paste0(readLines(filename), collapse = "\n")
```
G-Kazantzidis, line 225: I am thinking that maybe changing line in the end of the character string could be helpful for later `paste0(paste0(readLines(filename), collapse = "\n"), "\n")`. eg when we will add "}", we want them to be in an empty line at the end.

R/stan_module.R
G-Kazantzidis: apologies, I just saw that you create a new `is_file` function below

R/stan_module.R
gowerc: I think that would make more sense to go into the `as.character()` method no ?  As in (though unlikely) there is a chance this function could get used for other purposes where it doesn't make sense to have that trailing newline.

G-Kazantzidis: Your suggestion sounds more reasonable way to proceed.

G-Kazantzidis (approved):

PR #43, opened 2022-07-28
Add os model

### LINKED ISSUE #25: Create OSModel Object
- [ ] Constructor function
- [ ] parameterise / include_hazard method

G-Kazantzidis:
closes #25, #32

- `OsModel` class
- `LogLogisticOs` class 
- `LogLogisticModule` helper function (with default arguments for example Log-logistic model) 
- `LogLogisticOs` object creator 
- `parametrize` generic 
-  `parametrize` method for OsModel and HazardLink
- `parametrize` tests


R/OsModel.R
```diff
@@ -0,0 +1,171 @@
...
+            transformed_parameters = osmod@stan@transformed_parameters,
+            priors = osmod@stan@priors,
+            generated_quantities = osmod@stan@generated_quantities,
+            inits = osmod@stan@inits
+        ))
+
+        newOS@stan@functions <- str_replace_all(
+            string = osmod@stan@functions,
+            pattern = "<link_arguments>",
+            replacement = paste0("real ", link@parameters, ",")
```
gowerc, line 129: Should the "," be  ";" ?

R/OsModel.R
```diff
@@ -0,0 +1,171 @@
...
+    function(osmod, link) {
+        newOS <- OsModel(stan = StanModule(
+            functions = osmod@stan@functions,
+            data = osmod@stan@data,
+            parameters = osmod@stan@data,
+            transformed_parameters = osmod@stan@transformed_parameters,
+            priors = osmod@stan@priors,
+            generated_quantities = osmod@stan@generated_quantities,
+            inits = osmod@stan@inits
+        ))
```
gowerc, line 124: Not sure why you are creating a new StanModule argument when all the inputs are identical. Why not just have 

```
newOS <- osmod@stan
```

R/OsModel.R
```diff
@@ -0,0 +1,171 @@
...
+                replacement = paste0(link@parameters, link@contribution)
+            ) |>
+            str_replace(
+                pattern = "<link_arguments_as_par>",
+                replacement = paste0(link@parameters)
+            )
+
+
+        newOS@stan@priors <- append(osmod@stan@priors, link@stan@priors)
+        newOS@stan@inits <- append(osmod@stan@inits, link@stan@inits)
```
gowerc, line 143: Is this not combination not handled by the merge method ?

R/OsModel.R
```diff
@@ -0,0 +1,171 @@
...
+        )
+
+        newOS@stan@transformed_parameters <- str_replace(
+            string = osmod@stan@transformed_parameters,
+            pattern = "<link_log_surv>",
+            replacement = paste0(paste0(link@parameters, collapse = ","), ",")
+        ) |>
+            str_replace(
+
+                pattern = "<link_log_lik>",
```
gowerc, line 158:
```suggestion
            str_replace(
                pattern = "<link_log_lik>",
```

R/OsModel.R
```diff
@@ -0,0 +1,171 @@
...
+                replacement = paste0(paste0(link@parameters, collapse = ","), ",")
+            )
+
+        newOS@stan@generated_quantities <- str_replace_all(
+            string = osmod@stan@generated_quantities,
+            pattern = "<link_arguments_as_par>",
+            replacement = paste0(paste0(link@parameters, collapse = ","), ",")
```
gowerc, line 165: Should these be semi colons ?  i.e.  ";"

tests/testthat/test-osmodel.R
```diff
@@ -0,0 +1,29 @@
...
+        ),
+        parameters = "bla",
+        contribution = "bla"
+    )
+    os_mod_f <- parametrize(os_mod, h_link)
+
+    all(sapply(
+        c("link_arguments", "link_log_hazard_contribution", "link_arguments_as_par"),
+        grepl, os_mod_f@stan@functions
+    )) != TRUE
```
gowerc, line 22: We have used the testthat package here which means we need to use expectation functions i.e.  

```
expect_true(  1 ==1 )
```

Would recommend taking a look at: https://www.r-bloggers.com/2019/11/automated-testing-with-testthat-in-practice/

Am happy to have a chat to go over this if it would help

R/OsModel.R
```diff
@@ -0,0 +1,171 @@
...
+    )
+)
+
+
+#'  Log-logistic overall survival
+#'
+#' Description - Log-logistic overall survival model object creator
+#'
+#' @export
+LogLogisticOs <- setClass(
```
gowerc, line 22: Minor but perhaps worth moving this into its own file just to have separate between the classes

R/OsModel.R
```diff
@@ -0,0 +1,171 @@
...
+#' @param functions
+#' @param data
+#' @param parameters
+#' @param transformed_parameters
+#' @param generated_quantities
+#' @param priors
+#' @param inits
+#' @param templated
+#' @export
+LogLogisticModule <- function(functions = "os_functions.stan",
```
gowerc, line 40: I think these will need to point to the system files  as they won't be "local" when this is installed as a package

R/OsModel.R
G-Kazantzidis: Point. Thanks

R/OsModel.R
gowerc: Ah apoloiges I assumed (incorrectly) these were just individual statements as opposed to function inputs. Ignore me :)

R/OsModel.R
G-Kazantzidis: We can append the priors and the initial values lists using the merge method as well. I will implement it this way.

R/OsModel.R
G-Kazantzidis: This function should do the work, no ? 

https://github.com/Genentech/jmpost/blob/b69df60562ba6f07823b8218da88a59b2fc2b5cd/R/stan_module.R#L32-L42

tests/testthat/test-osmodel.R
G-Kazantzidis: I hope my last update fixed this issue.

gowerc:
@G-Kazantzidis , I think the error you are seeing is due to S4 being sensitive to the order in which the files are loaded. If you look in the `DESCRIPTION` file you will see there is a section for `collate` which specifies the order in which the files are loaded. I think you need to make sure that your OS model is loaded before your LogLogisticModel

R/OsModel.R
gowerc: O yes you are right apologies

R/OsModel.R
```diff
@@ -0,0 +1,127 @@
...
+    function(osmod, link) {
+        newOS <- osmod@stan
+
+        newOS@stan@functions <- str_replace_all(
+            string = osmod@stan@functions,
+            pattern = "<link_arguments>",
+            replacement = paste0("real ", link@parameters, ",")
+        ) |>
+            paste0("\\n ", link@stan@functions) |>
+            str_replace(
```
gowerc, line 88: Perhaps we can use `gsub()` instead of `str_replace()` to avoid the stringr dependency ?

R/LogLogisticOs.R
```diff
@@ -0,0 +1,37 @@
...
+setMethod(
+    f = "initialize",
+    signature = "LogLogisticOs",
+    definition = function(.Object,
+                          ..., stan = LogLogisticModule(), templated) {
+        callNextMethod(
+            .Object,
+            ...,
+            stan = stan@stan,
+            templated = stan@templated
```
gowerc, line 28: Sorry this is confusing me.  The `stan` arugment to initialise should be a `StanModule` object not an OS object. Also in this setup it looks like the `templated` argument is doing nothing ?

R/OsModel.R
```diff
@@ -0,0 +1,127 @@
...
+            transformed_parameters = transformed_parameters,
+            generated_quantities = generated_quantities,
+            priors = priors,
+            inits = inits
+        )
+
+    OsModel(
+        stan = st_mod,
+        templated = templated
+    )
```
gowerc, line 51: Ahh with regards to my above comment I think this function should just return the `StanModule` object not the OSModel object. This is just a help function for getting the correct stanmodule.  It does not need to be concerned with actually creating an OSModel

R/OsModel.R
```diff
@@ -0,0 +1,127 @@
...
+
+        newOS@stan@transformed_parameters <- str_replace(
+            string = osmod@stan@transformed_parameters,
+            pattern = "<link_log_surv>",
+            replacement = paste0(paste0(link@parameters, collapse = ","), ",")
+        ) |>
+            str_replace(
+                pattern = "<link_log_lik>",
+                replacement = paste0(paste0(link@parameters, collapse = ","), ",")
+            )
```
gowerc, line 116: I am wondering it if this can be simplified at all. Seems like quite a lot of repetition.

I'm wondering if we can defined some mapping list like:

```
list(
    "<link_log_surv>" = paste0(paste0(link@parameters, collapse = ","), ","),
    "<link_log_lik>" = paste0(paste0(link@parameters, collapse = ","), ","),
    "<link_parameters>" = link@stan@parameters
)

And then just loop over each element applying the replacements, rather than having to write them all individually out ? (I haven't looked into this in depth so apologies if this doesn't work.

```

tests/testthat/test-osmodel.R
```diff
@@ -0,0 +1,32 @@
...
+        ),
+        parameters = "bla",
+        contribution = "bla"
+    )
+    os_mod_f <- parametrize(os_mod, h_link)
+
+    expect_true(all(sapply(
+        c("link_arguments", "link_log_hazard_contribution", "link_arguments_as_par"),
+        grepl, os_mod_f@stan@functions
+    )) != TRUE)
```
gowerc, line 22: Instead of `!= TRUE` could you not just use `expect_false()` ?

tests/testthat/test-osmodel.R
```diff
@@ -0,0 +1,32 @@
+
+test_that("parametrize filled all the gaps", {
+    os_mod <- LogLogisticOs()
+    h_link <- HazardLink(
+        stan = StanModule(
+            functions = "bla",
+            data = "bla",
```
gowerc, line 7: Might be worth adding some more explicit tests to show that the merged stan objects come out as expected. I.e. after applying the character replacements  show that it equals a stanmodule built by hand with the replacements already applied.

tests/testthat/test-osmodel.R
```diff
@@ -0,0 +1,32 @@
+
+test_that("parametrize filled all the gaps", {
```
gowerc, line 2: Would also be good to add some tests to show the loglogistic model is picking up the code as expected.

R/LogLogisticOs.R
G-Kazantzidis: Regarding the Templated argument. Indeed, it has no role for now. I will delete it for now but here is the motivation behind. Before the OsModel is parametrized, it contains some place holders (eg <link_parameters>) that only exist to "save" space for the link parameters that will be selected later on. When the two sub-model are to be merged, we need to be sure that each and every place holder has been replaced with actual stan code. The templated slot would be `False` as long as there are still place holders in the stan code. It was supposed to become `True` only when the model is ready to be merged. So the idea was that I will work as an inhibitor for non - ready Os models.

R/OsModel.R
G-Kazantzidis: So if the `LogLogisticModule` produces a `StanModule` with default options for the Log-Logistic overall survival submodel then there is no need for the `LogLogisticOs` object creator, is that right ?

tests/testthat/test-osmodel.R
G-Kazantzidis: Yes. My mistake. Fixed

R/OsModel.R
G-Kazantzidis: Most likely it can be done more efficiently indeed.

R/OsModel.R
gowerc: I think we clarified this over the standup but just to clarify  that I would expect  `LogLogisticModule` to return a `StanModule` and then this would be the default input to the `LogLogisticOs` constructor

tests/testthat/test-osmodel.R
G-Kazantzidis: https://github.com/Genentech/jmpost/blob/0123fd37eb9c92d64ab2b941767822157607c02b/tests/testthat/test-osmodel.R#L72-L74

tests/testthat/test-osmodel.R
G-Kazantzidis: Like this one ? 
https://github.com/Genentech/jmpost/blob/0123fd37eb9c92d64ab2b941767822157607c02b/tests/testthat/test-osmodel.R#L36

R/LogLogisticOs.R
```diff
@@ -0,0 +1,36 @@
...
+#'
+#' Creates a log-logistic overall survival templated object
+#' @rdname LogLogisticOs-class
+#' @importFrom assertthat assert_that
+#' @export
+setMethod(
+    f = "initialize",
+    signature = "LogLogisticOs",
+    definition = function(.Object,
+                          ..., stan = LogLogisticModule()) {
```
danielinteractive, line 23: Do we even need to have this as an argument with default ? Would it not be better to just harcode the `stan = LogLogisticModule()` in line 27?

R/OsModel.R
```diff
@@ -0,0 +1,118 @@
+#' Title - TODO
+#'
+#' Description - TODO
+#'
+#' @slot stan TODO
```
danielinteractive, line 5: please complete

R/OsModel.R
```diff
@@ -0,0 +1,118 @@
...
+#' @slot templated Logical. Defines whether the OsModel object needs to be completed
```
danielinteractive, line 6:
```suggestion
```

R/OsModel.R
```diff
@@ -0,0 +1,118 @@
...
+#' @export
+OsModel <- setClass(
+    Class = "OsModel",
+    representation = list(
```
danielinteractive, line 10:
```suggestion
    slots = list(
```
See `?setClass`: "All these arguments are deprecated from version 3.0.0 of R and should be avoided."

R/OsModel.R
```diff
@@ -0,0 +1,118 @@
...
+        stan = "StanModule"
+    )
+)
+
+
+
+#' LogLogisticModule
+#'
+#' Log logistic module helper function.
```
danielinteractive, line 19:
```suggestion
#' Log logistic module helper function.
#'
```
similar in other places for visual readability

R/OsModel.R
```diff
@@ -0,0 +1,118 @@
...
+        inits = inits
+    )
+}
+
+
+
+
+
+
+
```
danielinteractive, line 54:
```suggestion
```

R/OsModel.R
```diff
@@ -0,0 +1,118 @@
...
+#' @param osmod TemplatedStanOs object
+#' @param link HazardLink object
+#' @export
+setGeneric("parametrize", function(osmod, link) {
+    standardGeneric("parametrize")
+})
+
+
+
+setMethod(
```
danielinteractive, line 65: where is this documented? I would suggest to use `@rdName parametrize` to put this in the same place as the generic function

R/OsModel.R
```diff
@@ -0,0 +1,118 @@
...
+                if (i == "functions" & names(gap_map)[k] == "<link_arguments>") {
+                    tmp_char <- paste0(tmp_char, "\n ", link@stan@functions)
+                }
+
+                slot(newOS@stan, i) <- tmp_char
+            }
+        }
+
+
+
```
danielinteractive, line 109:
```suggestion
```

R/OsModel.R
```diff
@@ -0,0 +1,118 @@
...
+        temp_obj <- merge(osmod@stan, link@stan)
+        newOS@stan@priors <- temp_obj@priors
+        newOS@stan@inits <- temp_obj@inits
+
+
```
danielinteractive, line 114:
```suggestion
```

inst/stanparts/os_functions.stan
```diff
@@ -0,0 +1,79 @@
...
+    for (i in 1:cols(times)) {
+      result[i] = times[i] < death_time;
+    }
+    return result;
+  }
+
+
+  row_vector get_cond_survival(real cens_surv_prob, row_vector uncond_surv_probs) {
+    row_vector[cols(uncond_surv_probs)] result = fmin(1.0, uncond_surv_probs / cens_surv_prob);
+    return result;
```
danielinteractive, line 78: all these functions are highly generic, they should not sit in model specific file but in general  `utils.stan` or so.

tests/testthat/test-osmodel.R
```diff
@@ -0,0 +1,107 @@
...
+        inits = list()
+    ))
+    expect_equal(actual, expected)
+})
+
+
+
+test_that("LogLogisticModule is working as expected", {
+    actual <- LogLogisticModule()
+    expected <- StanModule(
```
danielinteractive, line 74: I wonder if we should move towards using snapshots for testing, see https://testthat.r-lib.org/reference/expect_snapshot.html

danielinteractive (commented):
thanks a lot George! Please see comments inside

R/LogLogisticOs.R
G-Kazantzidis: `LogLogisticModule` is the core object containing all the stan code for the overall survival model. The user can either use our code or provide his own code (either by directly typing into R or as a .stan file). It should be as a default option for the case that the user wants to provide his of version of a loglogistic model. 
https://github.com/Genentech/jmpost/blob/1b4df12d55d50e10d79b99667a452db84efcd785/R/OsModel.R#L29-L35

inst/stanparts/os_functions.stan
G-Kazantzidis: I agree that we should better structure the "foldering" of the stan files. Separating the generic from the model specific stan parts already moves towards the direction of a more flexible package structure. Do those changes belong to this PR ? Or should we create a new branch specifically for this aim ?

R/LogLogisticOs.R
danielinteractive: Since the pattern is already in use in other places (e.g. exponential longitudinal model branch etc.) we keep this for now as is, potentially revisit in the future.

inst/stanparts/os_functions.stan
danielinteractive: handled in https://github.com/Genentech/jmpost/issues/68

inst/stanparts/os_functions.stan
```diff
@@ -0,0 +1,79 @@
+  matrix log_h0(matrix time, real lambda, real p) {
```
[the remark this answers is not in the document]
danielinteractive: please rename file

danielinteractive (approved):

Issue #38, opened 2022-07-22
Move assertions from constructor functions to validation functions


gowerc:
The below issue applies at the moment to the:
- [x] StanModule Object
- [x] HazardLink Object 

Currently all of our objects assertions are done in the constructor functions i.e.

```
setMethod(
    f = "initialize",
    signature = "StanModule",
    definition = function(
        .Object,
        ...,
    ) {
        assert_that(
            is.character(functions),
            is.character(data),
            is.character(parameters),
            is.character(transformed_parameters),
            is.character(generated_quantities),
            msg = paste(
                "`Functions`, `data`, `parameters`, `transformed_parameters` and",
                "`generated_quantities` must be character vectors"
            )
        )
```

I imagine some of these (though not all of them) should probabally be moved to the S4 validation function. From the [advanced R](https://adv-r.hadley.nz/s4.html) book:

![image](https://user-images.githubusercontent.com/8447209/180465594-efa1d13f-9618-416d-bd44-edeabc876a4a.png)

charksong:
In the character judgement of StanModule Object, I initially replaced _assert_that_ as

`setValidity("StanModule",function(object){
      if (!is.character(functions)
          |!is.character(data)
          |!is.character(parameters)
          |!is.character(transformed_parameters)
          |!is.character(generated_quantities)){
        "Functions, data, parameters, transformed_parameters and generated_quantities must be character vectors"
      }else{
        TRUE
      }
    })`

charksong:
By the way, scrolling up I noticed the parameter _representation_ is not recommended to write down in _setClass_ after version 3.0.0, should we enhance it or it does not matter?

![image](https://user-images.githubusercontent.com/56521967/181456093-b41ffbab-fed1-4ca0-b717-a8bed017f946.png)

The code is
```
StanModule <- setClass(
  "StanModule",
  representation(
    functions = "character",
    data = "character",
    parameters = "character",
    transformed_parameters = "character",
    priors = "list",
    generated_quantities = "character",
    inits = "list"
  )
)

```

PR #56, opened 2022-08-08
Validation Enhancement

charksong:
this solves #38


charksong:
I successfully passed the two checks, however, I noticed that all the codes were contributed by me from the github
image
Actually I just modified serveral lines and apparently not all of them (showing 77 and 248 additions relativelty), a silly question maybe but I am not sure that is normal. If normal, #38 should be finished

gowerc:
Hi @charksong , 

This issue appears to be that you've added the files to the root of the github directory:

image


The code should be in the R folder :)

Issue #57, opened 2022-08-09
Add slot "model" in the StanModule object. 


G-Kazantzidis:
Because of the special treatment required for the generation of the priors stan code in the section `model`, we introduce an additional slot in the `StanModule` object, named "model". This slot will be a character string which will include information relevant to the model stan code section. 

- [x] Prepare function that creates final character strings oriented by curly brackets `{...}`.
- [x] Create `StanModule` object with slots both `priors` and `model` where the former is a list and the later is a character. 
- [x] Add line  "target+=sum(log_lik);\n" at the end of the model section of the stan code. 
- [x] Update function `as.character`
- [x] Create function `model_prep` that prepares the model slot of the `StanModule` given the priors

PR #58, opened 2022-08-09
Add model slot

G-Kazantzidis:
#57


R/h_bracket.R
```diff
@@ -0,0 +1,21 @@
...
+#' @export
+setMethod(
+    f = "h_bracket",
+    signature = "list",
+    definition = function(x) {
+        model_text <- paste0(x, collapse = "\n")
+        paste0( model_text, "\n", "target+=sum(log_lik);\n")
+
+    }
+)
```
gowerc, line 21: Why does this need to be a S4 object and not just a  general function ?

R/stan_module.R
```diff
@@ -117,6 +124,18 @@ setMethod(
     }
 )
 
+#' model_prep. Populates the model section of a StanModule
+#' @param x A StanModule object
+#' @export
+model_prep <- function(x) {
+    if (any(is.na(x@model), nchar(x@model) < 2)) {
+        if (length(x@priors) > 0) x@priors <- as.list(paste(names(x@priors), "~", x@priors))
+        x@model <- h_bracket(x@priors)
```
gowerc, line 133: Do you not need to paste the model and the priors together ? If i'm reading this correctly it looks like the priors just overwrites the model ?

R/stan_module.R
```diff
@@ -117,6 +124,18 @@ setMethod(
     }
 )
 
+#' model_prep. Populates the model section of a StanModule
+#' @param x A StanModule object
+#' @export
+model_prep <- function(x) {
+    if (any(is.na(x@model), nchar(x@model) < 2)) {
```
gowerc, line 131: What about the case where model is populated ?  If i'm reading this right then it doesn't update the priors at all ? Do we not want to merge the priors in with the model ?

R/h_bracket.R
```diff
@@ -0,0 +1,21 @@
...
+
+#' @rdname h_bracket
+#' @export
+setMethod(
+    f = "h_bracket",
+    signature = "list",
+    definition = function(x) {
+        model_text <- paste0(x, collapse = "\n")
+        paste0( model_text, "\n", "target+=sum(log_lik);\n")
```
gowerc, line 18: How come we are implicitly adding  `target+=sum(log_lik)` shouldn't this be explicitly part of the model code in the OS model ?

R/h_bracket.R
G-Kazantzidis: I want the function to treat different list inputs (which is the prior lists) and character strings ( that is all the rest slots of StanModule). Thus, I need it to be generic. I am using the S4 method ( instead of S3) in order to be similar with the rest of the package.

R/stan_module.R
G-Kazantzidis: If the user provides a stan code for the model section, I do not want to overwrite it.

gowerc: But does this mean if model code is provided the priors is just ignored ? 

I guess I was thinking it would be something like (pseudo code):

```
if not missing(priors):
   model = combine(priors, model)
```

G-Kazantzidis: Well, I am not sure. I guess one of the two (priors or model) has to be ignored in the presence of the other. To my head, the user either specifies the prior list or he provides the complete stan code for the model section. I cannot see a way to "modify" a stan code given the prior list. I would assume that priority should be given to the stan code if provided.

R/h_bracket.R
G-Kazantzidis: Let me know If I understood your suggestion. The user will specify: 
`my_os_mod <- OsModel( stan = 
StanModule( model = "target+=sum(log_lik)" ))`. And then, when the Os and Long model will be merged, the final model section will use the prior list and combine it with the already existing model section code?

R/stan_module.R
gowerc: Hmmm I guess in my head I saw it working like this, like for example take the stan code:

```
model {
  sigma ~ lognormal(0, 0.5);         
  alpha ~ normal(0, sigma);       
  for (n in 1:N) {                        
    y[n] ~ bernoulli_logit(sum(alpha[team1[n]]) - sum(alpha[team0[n]]));
  }
}
```

Using our framework I imagine a valid way of splitting this code would be:

```
StanModule(
    priors = list(
        "sigma" = "lognormal(0, 0.5)",
        "alpha" = "normal(0, sigma)"
    ),
    model = c(
        "for (n in 1:N) {",
        "  y[n] ~ bernoulli_logit(sum(alpha[team1[n]]) - sum(alpha[team0[n]]));",
        "}"
    )
)
```

Likewise for the stanmodules which need that sum(log_lik) I imagine them being specified like:

```
StanModule(
    priors = list(
        "sigma" = "lognormal(0, 0.5)",
        "alpha" = "normal(0, sigma)"
    ),
    model = c(
        "target += sum(log_like)"
    )
)
```

As in the priors is an extension to the model which just provides some simple syntactic sugar

G-Kazantzidis: It seems more structured than my suggestion. I will implement it to see how it works out.

R/h_bracket.R
gowerc: urgh, yer fair enough. Personally I don't like S4 and try to use it as little as I can get away with but I appreciate thats purely a subjective argument and not based on anything realistic so am fine with this :)

R/h_bracket.R
gowerc: Not quite, I would expect the  `StanModule( model = "target+=sum(log_lik)" )` to be apart of the default Stan code that is returned by the `LogLogisticOs` object

R/stan_module.R
```diff
@@ -117,6 +124,16 @@ setMethod(
...
 )
 
+#' model_prep. Populates the model section of a StanModule
+#' @param x A StanModule object
+#' @export
+model_prep <- function(x) {
+    if (length(x@priors) > 0) x@priors <- as.list(paste(names(x@priors), "~", x@priors))
+    x@model <- paste(h_bracket(x@priors), x@model)
+    x
+}
```
gowerc, line 134: I think this is fine but can we include a unit test just to make sure we aren't modifying the original object i.e. if we did
```
y <- model_prep(x)
```

Making sure that `x` is unchanged. I am pretty sure this is how S4 works as I don't think it modifies in place but I just want to be sure.

R/h_bracket.R
```diff
@@ -0,0 +1,19 @@
...
+
+#' @rdname h_bracket
+#' @export
+setMethod(
+    f = "h_bracket",
+    signature = "list",
+    definition = function(x) {
+        paste0(paste0(x, collapse = "\n"), "\n")
+    }
+)
```
gowerc, line 19: If you are collapsing `priors` into strings and combining them with `model` is this function/method still needed?

tests/testthat/test-h_bracket.R
```diff
@@ -0,0 +1,20 @@
...
+
+
+test_that("h_bracket works as expected", {
+    st_mod <- StanModule(
+        functions = "string",
+        data = "string",
+        parameters = "string",
+        transformed_parameters = "string",
+        generated_quantities = "string",
+        priors = list("item1;", "item2;")
```
gowerc, line 11: Given the separation of `priors` and `model` now does it make sense to have unnamed priors anymore ? Perhaps we should throw an error for this ?

R/stan_module.R
```diff
@@ -117,6 +124,16 @@ setMethod(
     }
 )
 
+#' model_prep. Populates the model section of a StanModule
+#' @param x A StanModule object
+#' @export
+model_prep <- function(x) {
+    if (length(x@priors) > 0) x@priors <- as.list(paste(names(x@priors), "~", x@priors))
+    x@model <- paste(h_bracket(x@priors), x@model)
```
gowerc, line 132: Sorry am a bit lost here,  is this the right place to be calling `h_bracket()`  my understanding is that `h_bracket()` wraps the object in curly brackets `{  priors }`  but then you  are concatenating the model code on afterwards....  Shouldn't the model code and the priors be within the same set of curly brackets i.e.  `{ priors \n model }`

tests/testthat/test-h_bracket.R
G-Kazantzidis: Definitely, the priors ought to have names. I will make another test for the existence of the names of the priors.

R/stan_module.R
G-Kazantzidis: Correct. This is my motivation for making the function generic. When dealing with the priors, the function just creates a character text given the prior list but it does not add curly brackets. When dealing with characters, it adds the curly brackets.

R/h_bracket.R
G-Kazantzidis: I explain my motivation for the S4 further below.

R/stan_module.R
G-Kazantzidis: Let me know if I understood this correctly. Later on, when I invoke this function inside the `as.character`, I create a new object `y` instead of overwriting `x`. 

https://github.com/Genentech/jmpost/blob/bb2ddf7349021cfc308fb64f7121a15b45c7e3fb/R/stan_module.R#L146-L150

R/stan_module.R
gowerc: Ok yer I am not keen on this design of the functions.  The idea of generics is that you abstract away implementation details but that you should get consistent or similar outcomes.  Having the method  collapse if a list or  add brackets if a string are quite different outcomes and that can lead to confusion other the purpose of the method.

R/stan_module.R
gowerc: I just did some manual tests to confirm S4 doesn't  create references so yer this is all fine

tests/testthat/test-h_bracket.R
gowerc: Sorry did you add any assertions for this ?

G-Kazantzidis: I decided to ensure that priors have names with `assert_that` during the creation of the object in my todays commit .

R/stan_module.R
G-Kazantzidis: I changed the format from the S4 generic to simple function in the last commit.

tests/testthat/test-stanmodule.R
```diff
@@ -148,25 +154,9 @@ test_that("loading multiple lines from a file works as expected", {
...
-            "mystring4;",
-            "more strings",
-            "}",
-            ""
-        ),
-        collapse = "\n"
-    )
-    expect_equal(actual_char, expected)
-
-})
```
gowerc, line 169: @G-Kazantzidis , How come this was removed sorry ?

R/h_bracket.R
```diff
@@ -0,0 +1,16 @@
...
+#' @param x An Object
+#' @export
+h_bracket <- function(x) {
+    if (any(nchar(x) >= 1, length(x) > 1)) {
+        paste0(
+            "{\n",
+            paste0(x, collapse = "\n"),
+            "\n}\n"
+        )
+    }
```
gowerc, line 15: @G-Kazantzidis , Is this correct ? At the moment if  the `if()` condition is not met the function returns literally nothing ?

G-Kazantzidis: This is correct. The scenario that this is False is when we have a the character vector like `""`. In that case, the section is empty to there is no need to create it ( eg `generated quantities { }`.

G-Kazantzidis: However if the user provides a vector of two empty strings ( eg `stanmodule(model = c("","") )` ), it will return `"{\n \n}\n"`.

tests/testthat/test-stanmodule.R
G-Kazantzidis: Apologies for that. I might have deleted it by mistake when I changed the implementation of `as.character`. I put it back now.

R/stan_module.R
```diff
@@ -77,6 +78,13 @@ setMethod(
             msg = "`Priors` and `inits` must be lists"
         )
 
+        if (length(priors) > 0) {
+            assert_that(
+                is.null(names(priors)) == FALSE,
+                all(names(priors) != ""),
```
G-Kazantzidis, line 84: Here I ensure that all the names of the priors are not `NULL`.

R/stan_module.R
```diff
@@ -117,6 +131,29 @@ setMethod(
...
+    if (length(x@priors) > 0) {
+        x@priors <- as.list(
+            paste(names(x@priors), "~", x@priors)
+        )
+
+        x@model <- paste0(
+            paste0(x@priors, collapse = "\n"),
+            "\n",
+            x@model
+        )
```
G-Kazantzidis, line 150: I modified slightly the model prep function. I transfer the `x@model <- paste0(..` (line 146) inside the if loop because I want to modify the model section only in the case that priors are provided. If the priors are not provided, there is no need to change the model slot.

tests/testthat/test-stanmodule.R
```diff
@@ -149,11 +145,9 @@ test_that("loading multiple lines from a file works as expected", {
         c("mystring; mystring2\nmystring3;\nmystring4;", "more strings")
     )
 
-
     actual_char <- as.character(actual)
     expected <- paste0(
         c(
-            "",
```
G-Kazantzidis, line 156: Here, I deleted the `""` in line 156 because empty lines between sections are added only if we have more than one section. ( `paste(sections, collapse = "")`). If there is only one section slot, there are no additional empty lines.

R/h_bracket.R
gowerc: ah ok,  can we make this explicit though i.e.  adding a return outside of the `if()` condition that explicitly returns nothing.    `return(x)`  or  `return("")`  or  `return(invisible()`  at least this way it is clear it is a conscious decision. Also perhaps a unit test to test that this assumption is working as intended.

tests/testthat/test-stanmodule.R
gowerc: LGTM

tests/testthat/test-stanmodule.R
```diff
@@ -39,20 +38,17 @@ test_that("Loading from file works as expected", {
...
     )
 
     expected <- paste0(
         c(
-            "",
             "functions {",
             "some code",
             "some more code",
             "}",
-            "",
```
gowerc, line 55: @G-Kazantzidis ,

Did you mean to delete this one, based on your last comment I would have expected there to be a blank line here because its between 2 blocks ?

G-Kazantzidis: this is result of the `h_bracket` function 

https://github.com/Genentech/jmpost/blob/f8a2e4eaa49116fdd79530f519687da8a7605458/R/h_bracket.R#L13

gowerc: Sorry I'm not sure I follow surely we want a space between blocks i.e.

```
data {
code
}

functions {
code
}
```

As opposed to:

```
data {
code
}
functions {
code
}
```

R/stan_module.R
gowerc: lgtm

tests/testthat/test-stanmodule.R
G-Kazantzidis: I assume that this makes the stan code slightly more readable. Thus, I will go for the first option with the additional lines.

gowerc: yer apologies I appreciate this is pretty petty stuff now but is good to get it as nice as possible :)

R/h_bracket.R
G-Kazantzidis: The difference is that if we `return(x)` or `return("")`, an empty line is going to be added to the stan code. Whereas like this, no line is going to be added if a section slot is empty. I understand it is not a major difference.

tests/testthat/test-h_bracket.R
```diff
@@ -0,0 +1,27 @@
...
+    expect_equal(h_bracket(st_mod@generated_quantities), "{\nstring\n}\n")
+})
+
+
+test_that("Empty strings work fine", {
+    str <- ""
+    actual <- h_bracket(str)
+    expected <- NULL
+    expect_equal(actual, expected)
+})
```
G-Kazantzidis, line 27: this is the unit test to ensure that `h_bracket` works fine when the string is empty. Note that the result is different if the user provides a string with one empty space `" "`. I don't see why this will ever happen but if you consider it important I can add some code to avoid this case.

R/h_bracket.R
```diff
@@ -0,0 +1,23 @@
...
+            "\n}\n"
+        )
+
+    } else {
+
+        str <- NULL
+
+    }
+    return(str)
+}
```
G-Kazantzidis, line 23: I hope this is now closer to what you had in mind @gowerc

tests/testthat/test-h_bracket.R
gowerc: Might come back to this later but I think its fine for now to get this closed

gowerc (approved):

PR #60, opened 2022-08-10
Validation function modification

### LINKED ISSUE #38: Move assertions from constructor functions to validation functions

charksong:
closes #38


R/stan_module.R
```diff
@@ -41,8 +38,42 @@ read_stan <- function(string) {
...
+
+#' Validate Inputs
+#'
+#' @param object
+#'
+#' @return StanModule
+#' @export StanModule
+#'
+#' @examples StanModule
+#' @name setValidity2
```
gowerc, line 68: I don't think you need to provide a documentation file for this function

R/stan_module.R
```diff
@@ -41,8 +38,42 @@ read_stan <- function(string) {
...
+#'
+#' @examples StanModule
+#' @name setValidity
+setValidity("StanModule",function(object){
+    if (!is.character(object@functions)
+        |!is.character(object@data)
+        |!is.character(object@parameters)
+        |!is.character(object@transformed_parameters)
+        |!is.character(object@generated_quantities))
+        stop("`Functions`, `data`, `parameters`, `transformed_parameters`
```
gowerc, line 56: from the advanced R book ( [link](https://adv-r.hadley.nz/s4.html#validator) )  I think this function is supposed to return  TRUE (if it is valid) or a string (if it isn't valid). I don't think you are supposed to call the stop function yourself, the validation takes care of that for you behind the scenes

charksong: Hi Criag, by stop function I thought the codes may be more concise than the book's did
image
if stop function is redundant for some concern, i.e. computational cost, I will follow the book's default guide

R/hazard_link.R
```diff
@@ -20,34 +20,37 @@
...
+})
+
+setValidity("HazardLink", function(object) {
+    if (!length(object@contribution) == 1) {
+        "`contribution` must be length 1 character vectors"
+    } else {
+        TRUE
+    }
+})
+
```
gowerc, line 45: I think these overwrite each other,  I don't think they append for example:

```
myc <- setClass("myc", slots = list(x = "character"))

setValidity("myc", function(object) length(object@x) > 2)
setValidity("myc", function(object) length(object@x) == 1)

x <- myc(x = "a")
```

Doesn't error because the second condition takes precedence

gowerc:
@charksong ,  Lookings like the automated unit tests are failing due to the remove of `assertthat` from the imports. I think we need to list it as an import for some of the other functions for example I think it is used by `is_file()` in `stan_module.R`

NAMESPACE
```diff
@@ -17,3 +17,36 @@ exportMethods(getLinkDSLD)
...
 exportMethods(initialize)
 exportMethods(merge)
+<<<<<<< Updated upstream
+=======
+importFrom(assertthat,"TRUE")
+importFrom(assertthat,"for")
+importFrom(assertthat,"function")
+importFrom(assertthat,"if")
+importFrom(assertthat,A)
+importFrom(assertthat,Is)
```
gowerc: @charksong , Looks like something has gone wrong here. You've included the merge conflict  brackets (the `<<<<<<` and `=====` symbols)  and you also appear to have imported random words from the assertthat library that I'm pretty sure are not functions within it :)

R/hazard_link.R
```diff
@@ -44,6 +44,26 @@ setValidity("HazardLink", function(object) {
 })
 
 
+<<<<<<< Updated upstream
+=======
+setValidity("HazardLink", function(object) {
```
gowerc, line 49: You appear to have committed the merge conflict boundary flags

NAMESPACE
```diff
@@ -17,4 +17,33 @@ exportMethods(getLinkDSLD)
...
+importFrom(assertthat,or)
+importFrom(assertthat,return)
+importFrom(assertthat,short)
+importFrom(assertthat,string)
+importFrom(assertthat,that)
+importFrom(assertthat,to)
+importFrom(assertthat,utility)
+importFrom(assertthat,valid)
+importFrom(assertthat,well)
+importFrom(assertthat,will)
```
gowerc, line 49: @charksong , Looks like something is going wrong with your documentation as it appears to have added a whole sentence to the NAMESPACE file rather than specific functions.  At least I'm pretty sure there is no `will` function coming from the `assertthat` library :)

R/stan_module.R
```diff
@@ -213,12 +187,13 @@ read_file <- function(filename) {
 }
 
 
+#' @importFrom assertthat assert_that
```
gowerc, line 190: @charksong  I think the issue I mentioned above about the NAMESPACE file is coming from this line. I think this needs to be placed below the function description

charksong, line 190: Hi Craig @gowerc, though the checks passed however I imported random words from the assertthat library, I think the problem is from my moved `#' @importFrom assertthat assert_that` on line 45 before S4 function directly to that before S3 function `is_file`, and it's a matter of roxygen2's syntax. If my guess is right, I'm going to read carefully roxygen2's guide.

charksong:
Thanks a lot Craig! I think I'm going to learn more about roxygen2's syntax.

charksong:
Hi Craig@gowerc , I think this finally may be closed. Thanks for review.

R/hazard_link.R
```diff
@@ -20,34 +20,32 @@
...
+        stan = "StanModule",
+        contribution = "character",
+        parameters = "character"
     )
 )
 
+setValidity("HazardLink", function(object) {
+    if (!length(object@parameters) >= 1) {
+        "`parameter` must be a vector"
+    }else if(!length(object@contribution) == 1){
```
danielinteractive, line 33: same here

danielinteractive (commented):
thanks Marco, please see inside

danielinteractive (approved):
thanks Marco!

danielinteractive:
@charksong I see that we got errors on the checks, I guess we need a new issue that addresses and fixes those .
please next time don't merge when there are check errors.
@gowerc fyi - maybe it would be better to force check passes and otherwise don't allow merges
image

charksong:
@danielinteractive @gowerc 
My apologies, maybe we could apply `git reflog` and `git reset --hard 6bcedd2e7d7d399e3aaccac52f29ca7651036b44` to back on the last version before this merging by George.

danielinteractive:
@charksong hm I think the merge was done by you, not George - but does not matter. I would not reset for this one, just fix the problem is sufficient

gowerc:
@danielinteractive , I had enabled that option. Not sure how this still went through :S

image

I think from the log it looks like you had approved it ?

danielinteractive:
Yes I approved it because I assume checks will be and need to be fixed

danielinteractive:
@cicdguy do you have any idea? we are out of ideas :-)

PR #69, opened 2022-09-08
Add JmModel object 

### LINKED ISSUE #28: Create JointModel Object
- [x] Constructor function
- [x] compile method

G-Kazantzidis:
closes #28 

Issues 
- Test for miss specified stan code aims to check if the stan files have been loaded correctly to the `LongModel` or `OsModel` objects. It does that by searching for patterns "/.stan". The idea is that when the constructor `StanModule` can be populated either by a stan file `StanModule( functions = "file.stan")` or with character strings `StanModule( functions = "code;")` . First the constructor searches if the provided character is a file inside the `jmpost` package directory. I there is no such file, the character is assumed to be code. In our case, although all the .stan files have been correctly loaded to the constructor, the pattern ".stan" is still found in the output. 
https://github.com/Genentech/jmpost/blob/14fd25dc4b0f61b97da5dbfae9eef9a2c5847615/tests/testthat/test-JmModel.R#L21


NAMESPACE
```diff
@@ -21,10 +22,14 @@ exportClasses(LogLogisticOs)
 exportClasses(LongModel)
 exportClasses(OsModel)
 exportClasses(StanModule)
+exportClasses(jmModel)
```
danielinteractive, line 25: better name would be `JointModel` because `jm` is already abbreviating "Joint Model" and the others end with "Model"

NAMESPACE
```diff
@@ -5,11 +5,12 @@ export(LogLogisticModule)
 export(LogLogisticOs)
 export(LongModel)
 export(OsModel)
-export(StanModule)
 export(getLink)
 export(getLinkDSLD)
 export(getLinkTTG)
 export(h_bracket)
+export(jm_complete)
```
danielinteractive, line 12: better name could be just `joint()`?

R/JMModel.R
```diff
@@ -0,0 +1,42 @@
...
+#' @export
+setMethod(
+    f = "initialize",
+    signature = "jmModel",
+    definition = function(.Object,
+                          ...,
+                          Long,
+                          Os,
+                          name = "stan_model.stan") {
+        mod <- jm_complete(Long = Long, Os = Os)
```
danielinteractive, line 23: I think this could read nicely with `joint(long = long, os = os)` - I would avoid the uppercase argument names

R/jm_complete.R
```diff
@@ -0,0 +1,39 @@
...
+#' @rdname jm_complete
+setMethod("jm_complete",
+    signature(Long = "LongModel", Os = "OsModel"),
+    value = "StanModule",
+    def = function(Long, Os) {
+        priors <- append(Long@stan@priors, Os@stan@priors)
+        priors <- priors[!duplicated(names(priors))]
+
+        StanModule(
+            functions = paste0(Long@stan@functions,
```
danielinteractive, line 11: I thought we have a general method to combine two stan objects together that would take care behind the scenes of combining all of this?

inst/LogLogistic/exp_long_functions.stan
```diff
@@ -0,0 +1,101 @@
+row_vector ifelse(int[] condition, row_vector yes, row_vector no) {
```
danielinteractive, line 1: I think it would now be already a good time to make a `utils.stan` file which is appended to the functions when doing the `complete()` call.

danielinteractive (commented):
thanks George, good work, please see first comments inside

NAMESPACE
G-Kazantzidis: Nice observation. I will fix it.

NAMESPACE
G-Kazantzidis: Done

R/jm_complete.R
G-Kazantzidis: This are specifically a `LongModel` and an `OsModel` though. They need to be treated in a special way compared to other cases.

R/JMModel.R
G-Kazantzidis: fixed

inst/LogLogistic/exp_long_functions.stan
G-Kazantzidis: I will implement it into this branch then

R/jm_complete.R
danielinteractive: Interesting. In that case I would propose to write a method that works for these two - then the code is cleaner here / easier to read. This works because S4 can dispatch on multiple arguments, so when it sees exactly this combination then it will use the special method

G-Kazantzidis: I am not sure I understand. I am using an S4 method (`setMethod`) with specific signatures for `LongModel` and `OsModel`.
https://github.com/Genentech/jmpost/blob/043e99a20f96307c6beb99d78292a9c86bc7746e/R/joint.R#L4
Is this different from your suggestion ?

G-Kazantzidis: Maybe also an additional sidenote: `StanModule` is the core object that contains different parts of stan code. All objects contain it (`OsModel`, `LongModel` etc). I am referring to this case as "special" because it is different from the `parametrize` for example ( where we combine an `OsModel` with the link but practically the output is again a new `StanModule`).

danielinteractive: yeah it would be like here, but just to combine the two objects. probably I am just proposing to not define a new generic function `jm_complete()` but instead just define a method for the existing generic that combines two general stan objects.

G-Kazantzidis: Daniel, you can now find the updated generic `merge` method including also what was previously known as `Joint` ( and even more previously as `jm_complete`). `merge` was initially constructed for merging object of identical class. I had to drop this restriction in order to include the merging of the longitudinal and overall survival model.

danielinteractive: thanks George, sounds good

inst/LogLogistic/exp_long_functions.stan
danielinteractive: that is a good idea I think! Because `as.character()` creates the final stan program for any `StanModule`.

danielinteractive (commented):
please see my comments, please re-request review once ready, thanks Geroge

R/JointModel.R
```diff
@@ -0,0 +1,48 @@
...
+    signature = "JointModel",
+    definition = function(.Object,
+                          ...,
+                          long,
+                          os,
+                          util = TRUE,
+                          name = "stan_model.stan") {
+        mod <- merge(x = long, y = os)
+
+        utils <- StanModule(functions = read_file(system.file("Utils", "utils.stan", package = "jmpost")))
```
G-Kazantzidis, line 27: Maybe we should consider making the `read_stan` function more generic. Right now it only can search files into the LogLogistic folder of the package.

inst/LogLogistic/exp_long_functions.stan
G-Kazantzidis: You can find my implementation in the new update

R/JointModel.R
danielinteractive: yes please , that should be more generic

danielinteractive: maybe do a separate issue and PR for that though

danielinteractive (approved):
thanks @G-Kazantzidis , ok to merge and the continue with improvement of read_utils etc. in separate issues/PRs.

Issue #70, opened 2022-09-12
read_stan generic 


G-Kazantzidis:
Make `read_stan` function more flexible in order to be able to search for .stan file in multiple locations in the `jmpost` object. 
Notes: 

- When merging the utils, general functions **must** go at the beginning.

PR #71, opened 2022-09-13
Update read_stan 

G-Kazantzidis:
#70 The aim of the branch is to allow for the `read_stan` function to read files from multiple folders inside the `jmpost` directory.

### DIFF
--- a/R/JointModel.R
+++ b/R/JointModel.R
@@ -24,9 +24,9 @@ setMethod(
                           name = "stan_model.stan") {
         mod <- merge(x = long, y = os)
 
-        utils <- StanModule(functions = read_file(system.file("Utils", "utils.stan", package = "jmpost")))
+        utils <- StanModule(functions = "utils.stan")
 
-        if(util == TRUE) mod <- merge(x = mod, y = utils)
+        if(util == TRUE) mod <- merge(x = utils, y = mod)
 
         cmdstanr::write_stan_file(
             as.character(mod),
--- a/R/JointModel.R
+++ b/R/JointModel.R
@@ -26,7 +26,7 @@ setMethod(
 
         utils <- StanModule(functions = "utils.stan")
 
-        if(util == TRUE) mod <- merge(x = utils, y = mod)
+        if(util) mod <- merge(x = utils, y = mod)
 
         cmdstanr::write_stan_file(
             as.character(mod),
--- a/R/stan_module.R
+++ b/R/stan_module.R
@@ -38,7 +38,16 @@ read_stan <- function(string) {
         package = "jmpost",
         string = string
     )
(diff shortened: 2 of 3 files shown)


G-Kazantzidis:
@danielinteractive We should chat about an issue with the snapshot tests. Something seems to be wrong there.

R/JointModel.R
```diff
@@ -24,9 +24,9 @@ setMethod(
                           name = "stan_model.stan") {
         mod <- merge(x = long, y = os)
 
-        utils <- StanModule(functions = read_file(system.file("Utils", "utils.stan", package = "jmpost")))
+        utils <- StanModule(functions = "utils.stan")
 
-        if(util == TRUE) mod <- merge(x = mod, y = utils)
+        if(util == TRUE) mod <- merge(x = utils, y = mod)
```
danielinteractive, line 29:
```suggestion
        if(util) mod <- merge(x = utils, y = mod)
```
generally if `util` is a flag then just do like this - if not sure then assert before

R/stan_module.R
```diff
@@ -31,7 +31,18 @@ StanModule <- setClass(
...
-    system_file <- system.file("LogLogistic", string, package = "jmpost")
+    file_map <- list.files("inst/")
+    system_file <- apply(as.matrix(file_map),
+        1,
+        FUN = system.file,
+        package = "jmpost",
+        string = string
+    )
+    system_file <- system_file[which(nchar(system_file) != 0)]
+    if (length(system_file) == 0 || length(system_file) > 1 ) system_file <- ""
```
danielinteractive, line 42: why in the case of more than one file found we don't use them?

danielinteractive (commented):
thanks George, please see inside one question

R/stan_module.R
G-Kazantzidis: It refers mainly to the case where the user leaves a section empty (eg the model section here `StanModule(functions = "file.stan", model = "")`). In that case, for the model section, the call is `read_stan("")` and therefore at some point inside the `read_stan` function we have the call `system_file( "folder_name", "", package = "jmpost")`.  The last call gives back the path for the folder. Since we have 3 folders inside the `inst` directory, we get back 3 characters with the 3 different paths.

danielinteractive (approved):

R/stan_module.R
```diff
@@ -31,7 +31,27 @@ StanModule <- setClass(
 #' file in the package directory or the stan code as a string.
 #' @export
 read_stan <- function(string) {
-    system_file <- system.file("LogLogistic", string, package = "jmpost")
+    file_map <- list.files("inst/")
```
danielinteractive, line 34: @gowerc need to fix this, sorry

PR #75, opened 2022-09-23
add getter and setter method for priors

### LINKED ISSUE #61: Add `priors()`  getter and setter method
To ease the burden of changing priors on existing objects it would be good to have a `priors()`  method for `stanmodule`, `hazardlink` , `longmodel` and `osmodel`  that can both access the existing defined priors and set them. Ideally so something like the following works:

```r
lm <- ExpLongModel(StanModule(...))
link <- get_link(lm, selected = c("ttg"))
x <- priors(link)
x$beta_ttg <- "normal(0,1)"
priors(link) <- x
```

if this works, automatically the following will also work:

```r
priors(link)$beta_ttg <- "normal(0,1)"
```
and this only needs 1 LoC instead of 3 LoC above.

Also this will work:
```r
priors(link)[["logit(mu_phi[sld_par_shared])"]] <- "normal(0,1)"
```

Use the following as starting point / pattern:

```r
.A <- setClass("A", slots = list(x = "numeric", y = "logical"))

a <- .A(x = c(1, 5), y = TRUE)

a

# getter method for x:
setGeneric(name = "x", def = function(object) standardGeneric("x"))

setMethod(
  f = "x", 
  signature = list(object = "A"),
  definition = function(object) {
    object@x
  }
)

x(a)

# setter method for x:
setGeneric(name = "x<-", def = function(object, value) standardGeneric("x<-"))

setReplaceMethod(
  f = "x",
  signature = "A",
  definition = function(object, value) {
    object@x <- value
    object
  }
)

x(a) <- c(5, 6)
x(a)
a

x(a)[1] <- 1
x(a)
```

shawnafang:
closes #61 

add getter and setter method for priors for long model, stan, OS model and hazard link


R/setPrior.R
```diff
@@ -0,0 +1,105 @@
...
+
+#method for stanmodule, hazardlink , longmodel and osmodel
+
+# setMethod(
+#     f = "[[<-",
+#     signature = c("PersonList"),
+#     definition=function(x,i,j,value) {
+#         x@datasetList[[i]] <- value
+#         return(x)
+#     })
```
danielinteractive, line 105: please remove commented code from PR
```suggestion
```

R/setPrior.R
```diff
@@ -0,0 +1,105 @@
+
+#priors() method for stanmodule, hazardlink , longmodel and osmodel
+
+# getter method for LongModel:
+setGeneric(name = "prior", def = function(object) standardGeneric("prior"))
```
danielinteractive, line 5: please add generic definitions to `generics.R` instead of here

R/setPrior.R
```diff
@@ -0,0 +1,105 @@
...
+    signature = "HazardLink",
+    definition = function(object) {
+        object@stan@priors
+    }
+)
+
+#' get method for stan module
+
+
+setMethod(
```
danielinteractive, line 38: please move the `StanModule` methods to the file `stan_module.R` - sorry I think I said something else in the last standup, but looking at how the other code is structured that makes more sense to keep the methods with the classes.

R/setPrior.R
danielinteractive: also please use plural form `priors` for all generics and methods

R/setPrior.R
```diff
@@ -0,0 +1,105 @@
...
+#priors() method for stanmodule, hazardlink , longmodel and osmodel
+
+# getter method for LongModel:
+setGeneric(name = "prior", def = function(object) standardGeneric("prior"))
+#' get method for longmodel
+setMethod(
+    f = "prior",
+    signature = list(object = "LongModel"),
+    definition = function(object) {
+        object@stan@priors
```
danielinteractive, line 11: for all classes which are not `StanModule` I would reuse the method, i.e.
```suggestion
        priors(object@stan)
```
this simplifies maintenance - if we decide later to store the priors differently within `StanModule` then we only need to change that single method and not everywhere

R/setPrior.R
```diff
@@ -0,0 +1,105 @@
+
+#priors() method for stanmodule, hazardlink , longmodel and osmodel
+
+# getter method for LongModel:
+setGeneric(name = "prior", def = function(object) standardGeneric("prior"))
+#' get method for longmodel
+setMethod(
+    f = "prior",
+    signature = list(object = "LongModel"),
```
danielinteractive, line 9: also please move the `LongModel` methods to the `longmodel.R` file etc.

tests/testthat/test-setPrior.R
```diff
@@ -0,0 +1,81 @@
+test_that("Priors of the stanmodule is replaced ", {
```
danielinteractive, line 1: please move that to the stanmodule test file

tests/testthat/test-setPrior.R
```diff
@@ -0,0 +1,81 @@
...
+
+    prior(stan)["mu_kg"]<-"lognormal(-0.24,1);"
+    actual_replaced<- prior(stan)["mu_kg"]
+    expected_replaced<-list(mu_kg="lognormal(-0.24,1);")
+    expect_equal(actual_replaced, expected_replaced)
+
+})
+
+
+test_that("Priors of the Long model is replaced ", {
```
danielinteractive, line 20: similarly move this to longmodel test file etc.

danielinteractive (commented):
thanks @shawnafang , nice work! apologies I think I sent you a bit in the wrong direction, I think now that it is better to have this distributed among files, see my comments. but that should not take long to move things around.

R/setPrior.R
shawnafang: I noticed this right after merged. Just moved

R/setPrior.R
shawnafang: Sure! Also move the generic definitions of setGeneric(name = "priors<-",...) to generics.R, right?

R/setPrior.R
shawnafang: Thanks for the suggestion! I will modify accordingly.

R/setPrior.R
danielinteractive: yep exactly :-)

R/generics.R
```diff
@@ -20,3 +20,13 @@ setGeneric(
 setGeneric("parametrize", function(osmod, link) {
     standardGeneric("parametrize")
 })
+
+
+#‘ priors() method for stanmodule, hazardlink , longmodel and osmodel
+
```
danielinteractive, line 26: please follow same documentation pattern as above

R/generics.R
```diff
@@ -20,3 +20,13 @@ setGeneric(
...
+setGeneric(name = "priors", def = function(object) standardGeneric("priors"))
+
+#‘ priors() replacement method for
```
danielinteractive, line 29: same here

tests/testthat/test-hazardlink.R
```diff
@@ -49,3 +49,30 @@ test_that("HazardLinks can merge together", {
     expect_equal(actual@stan@parameters, c("real beta_dsld;", "real beta_ttg;"))
 
 })
+
+
+
+test_that("Priors of the Hazardlink is replaced ", {
+    hazlink<-HazardLink(
```
danielinteractive, line 56: please install `styler` package and style the file or at least the new code section

danielinteractive (commented):
thanks @shawnafang , looks better now, please see additional comments

R/OsModel.R
```diff
@@ -108,6 +108,12 @@ setMethod(
 
 #' getter method for OsModel
 
+#' getter method for OsModel:
+#' @rdname priors
+#' @param object A `OsModel` object generated from function [OsModel()]
+#' @export
+
+
```
danielinteractive, line 116:
```suggestion
#' @rdname priors
#' @export
```
is actually sufficient and better because the general `priors()` documentation is otherwise overwritten. please use same pattern for the other methods

R/generics.R
```diff
@@ -22,11 +22,31 @@ setGeneric("parametrize", function(osmod, link) {
...
 
-#‘ priors() replacement method for
-setGeneric(name = "priors<-", def = function(object, value) standardGeneric("priors<-"))
 
+#' priors() get method for `LongModel`, `HazardLink` , `StanModule` or `OsModel`
+#' @rdname priors
+#' @param object A `LongModel`, `HazardLink` , `StanModule` or `OsModel` object
+#' @examples
+#' priors(object)
+#' @export
```
danielinteractive, line 33: exactly, here we need it because we document the generic function. so keep this

R/generics.R
```diff
@@ -22,11 +22,31 @@ setGeneric("parametrize", function(osmod, link) {
...
+)
 
+#' priors() replacement method for `LongModel`, `HazardLink` , `StanModule` or `OsModel` object
+#' @rdname extract-priors
+#' @param object A `LongModel`, `HazardLink` , `StanModule` and `OsModel` object
+#' @param value the character strings of the prior information for replacement
+#' @examples
+#' priors(object)<-"lognormal(0,1);"
+#' @export
+
```
danielinteractive, line 47: also keep this
```suggestion
#' priors() replacement method for `LongModel`, `HazardLink` , `StanModule` or `OsModel` object
#' @rdname extract-priors
#' @param object A `LongModel`, `HazardLink` , `StanModule` and `OsModel` object
#' @param value the character strings of the prior information for replacement
#' @examples
#' priors(object)<-"lognormal(0,1);"
#' @export
```
just no blank line

danielinteractive (commented):
thanks @shawnafang please see last comments

R/generics.R
shawnafang: Thank you and will change accordingly.

danielinteractive (approved):

PR #76, opened 2022-10-05
Fix the bug for S3_assertthat to S4_validation

charksong:
This solves the bug in #60


charksong:
@danielinteractive Sorry for the confusion, I made update to fix that test failure neglected last merging, and it's ready to be merged now.

danielinteractive (commented):
Thanks for working on this, but the fix should still use validation instead of assertion

R/stan_module.R
```diff
@@ -21,6 +21,17 @@ StanModule <- setClass(
...
+setValidity("StanModule", function(object) {
+    msg <- NULL
+
+    if (!is.scalar(object@priors) && is.null(names(object@priors))) {
+        msg <- c(msg, "`Priors` must have names")
+    }
+
+    msg
+})
+
```
charksong, line 35: I believe those check errors are from R codes here, and almost all of them reported the same error info as follows
image

charksong: If these functions were coded by assertion, it will be fine. 
`if (length(priors) > 0) {`
` assert_that(`
   `is.null(names(priors)) == FALSE,`
       ` all(names(priors) != ""),`
       ` msg = "Priors must have names"`
`    )`
`}` However, I cannot really understand the functional difference between my new code and the assertion code before.

danielinteractive: So when I am looking at the `R CMD check` errors, most of them come from an object where the priors does not have names. I think you can start from those tests/errors and see where we are missing the prior names, and then we add them there?
@gowerc fyi

danielinteractive:
@charksong this is a priority PR for us, can you still finish this? or should somebody else take over?

charksong:
> @charksong this is a priority PR for us, can you still finish this? or should somebody else take over?

Sorry, I am stuck by some other tasks, and I have not located where prior names missing. If somebody can take it over I would be appreciated, also I am curious and willing to learn from his/her codes for this modification. Thanks.

danielinteractive:
thanks @charksong , so @gowerc can take over

R/h_bracket.R
```diff
@@ -24,5 +24,5 @@ h_bracket <- function(x) {
 str_map <- function(i) {
     i <- gsub("^\\s*", "", i)
     i <- gsub("\n\\s*", "\n    ", i)
```
danielinteractive, line 26:
```suggestion
    gsub("\n\\s*", "\n    ", i)
```
without `return` below would be standard (avoiding explicit `return`s)

danielinteractive (approved):

PR #79, opened 2022-10-21
Add validationcheck prior

shawnafang:
to check if the replaced priors fits the corresponding model.


shawnafang:
Seems the priors of the other test function must be added?

NAMESPACE
```diff
@@ -1,6 +1,7 @@
 # Generated by roxygen2: do not edit by hand
 
 export("priors<-")
+export()
```
gowerc, line 4: Not quite sure where this is coming from but it indicates something isn't correct as none of the exports should be blank

shawnafang: >is this relavent with the warning of :
```r
Warning messages:
1: [hazard_link.R:98] Block must have a @name
i Either document an existing object or manually specify with @name 
```

gowerc: O yer that would be the one, `@name` should always have an argument. To be honest I don't think it is needed I think you can just delete it.

shawnafang:
Will pull again when the target priors for specific long model and os model is updated. 
validation function probabally should be on ExponentialLongModel .
os_priors should be placed onto the LogLogisticOSmodel not the OSmodel

PR #134, opened 2023-05-10
Use IDR workflows

cicdguy:
Instead of the custom ones.


danielinteractive (approved):
thx a lot Dinakar! much appreciated

danielinteractive:
Thanks @cicdguy , let me know once you are out of luck, then I can push more fixes on this branch, so that we avoid broken `main`

danielinteractive:
Maybe we can just set a linter / styler option for 4 spaces indentation? If not, I would suggest to go with the standard = supported by the linter and styler packages

danielinteractive:
Could this be helpful?
https://community.rstudio.com/t/change-default-indent-in-styler-to-4-spaces-vs-2/18474

danielinteractive:
@gowerc Can you make the call as Eng lead here regarding the trade off between better readability for you and maybe others vs higher maintenance effort? (custom workflow, editing settings, difference from other packages)

danielinteractive:
Thanks @gowerc, ok then let's go for 4 spaces. @cicdguy

danielinteractive:
@cicdguy Is my understanding correct that we basically just take the default `style.yaml` from the template but then modify line https://github.com/insightsengineering/r.pkg.template/blob/main/.github/workflows/style.yaml#L82 to say 
`styler::style_file(changed_r_files, style = tidyverse_style(indent_by = 4L), dry = dry)`
?

danielinteractive:
Or, alternatively, could you add a number parameter to your template workflow, that can then be modified in downstream repos? e.g. `indent_by` and then we can just have a `yaml` parameter changed to 4 (from original default 2)?

danielinteractive:
@cicdguy I just tried to resolve the merge conflicts but that seems too complicated, I would suggest to close this PR and make a new one, including the indentation setting adaptation as discussed above

PR #158, opened 2023-06-07
Fixes for GSF simulation

### LINKED ISSUE #123: GSF model converges but has poor diagnostic properties
Currently the GSF model is able to converge to known "real" parameters from a simulated dataset however it is displaying poor diagnostics which give the impression the model is not healthy (e.g. low ESS, high sample correlations).

gravesti:
Fixes #123 

This bug caused a significant part of this issue. Otherwise the convergence is greatly impacted by the simulation parameters, initial conditions and warm up. A long warmup fixes the warnings about treedepth and divergence.


danielinteractive:
Thanks @gravesti , I think some tests need to be adapted, but let's have a look tomorrow at that

gravesti:
We can discuss if these changes make sense or should go in a separate PR. Currently survival events can only occur at `times` specified in `simulate_joint_data()`.

PR #188, opened 2023-08-05
House Keeping

gowerc:
General code re-organisation / local test updating

- added keri test
- updated tests for current api
- Split simulations into multiple files

Accidentally included  `R/simulate_sf.R` from another branch. Its just 1 file though so shouldn't be to much of an issue....

### DIFF
--- a/design/tests/gsf-no-link.R
+++ b/design/tests/gsf-no-link.R
@@ -132,13 +132,15 @@ jdat <- DataJoint(
         formula = Surv(time, event) ~ cov_cat + cov_cont,
         subject = "pt",
         arm = "arm",
-        study = "study"
+        study = "study",
+        time_grid = c(1)
     ),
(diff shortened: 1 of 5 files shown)


DESCRIPTION
```diff
@@ -79,5 +79,9 @@ Collate:
     'argument_convention.R'
     'defaults.R'
     'jmpost-package.R'
+    'simulate_sf.R'
```
danielinteractive, line 82: Should this be called instead `simulations_sf.R`?

danielinteractive (dismissed):

DESCRIPTION
danielinteractive: Ah I see, if you merge this PR first and the SF PR later then it should be all fine moving this file to the SF branch

danielinteractive (approved):

PR #193, opened 2023-08-07
Feature/zero covariates

### LINKED ISSUE #190: Package doesn't support 0 covariates on OS model
Currently if you specify just an intercept (or more appropriately no covariates) on the OS model then the package just errors:

```
 p_os_cov_design is 0, but must be greater than or equal to 1.000000
```

gowerc:
Closes #190

- Updated Stan code for it to accept 0 covariate parameters. Wasn't confident about how stan behaves with a 0 dimension matrix (and didn't want to rely on implicit behaviour) so implemented this by checking to see if `p==0` and then creating a vector with all elements set to 0.

- Extracted the covariate contribution logic into its own function for clarity 

- Tested this by adding a call to the exponential model both with and without covariates.  Test passes if the z-score of the main parameters is within an acceptable margin of error

- Added a model storage directory `MODEL_DIR` in anticipation of https://github.com/insightsengineering/idr-tasks/issues/628

### DIFF
--- a/inst/stan/base/survival.stan
+++ b/inst/stan/base/survival.stan
@@ -59,7 +59,7 @@ functions {
     }
     
     
-    // Exand design matrix by parameter vector
+    // Expand design matrix by parameter vector
     vector get_os_cov_contribution(matrix design, vector beta) {
         int n = rows(design);
         int p = cols(design);


inst/stan/base/survival.stan
```diff
@@ -57,6 +57,26 @@ functions {
...
+    
+    
+    // Expand design matrix by parameter vector
+    vector get_os_cov_contribution(matrix design, vector beta) {
+        int n = rows(design);
+        int p = cols(design);
+        vector[n] cov_contribution;
+        if (p == 0) {
+            // 0 Vector to cover case where no covariates are defined
+            // (e.g. intercept only model)
```
danielinteractive, line 69: Just to confirm: normally the design matrix we use for the covariate contribution would not contain a column of 1s, ie corresponding to an intercept?

danielinteractive: yeah and in the code itself do we remove the 1s column of the design matrix before passing it to Stan?

danielinteractive: ok all good then

danielinteractive (approved):

PR #203, opened 2023-08-23
Move generated quantities to separate post sampling call

### LINKED ISSUE #181: Move survival curve + longitudinal prediction curves to post processing
Currently the fitted values for the survival and longitudinal plots are done in stan's generated quantities. These are implemented for all patients and are done for multiple time points. This results in a parameter explosion that significantly decreases the performance of querying the stan object (e.g. running `summary()`). 

Given that the user may want to select different time_grids or only sample a subset of patients it seems sensible to me that we implement this in post-processing in R instead.

gowerc:
Closes #181 

@danielinteractive  - I have marked as draft as this is not yet complete but was hoping to get initial feedback before I spend time updating all the unit tests and documentation.

Main changes are:

- The generated quantities are now created by a separate call to `generateQuantities(samples)`
- `time_grid` is no longer passed to the `Data*` objects but is instead an argument to `generateQuantities`
- The `longitudinal` and `survival` plotting functions now work off of this object
- The PR looks worse than it is because I moved `longitudinal` and `survival` from `R/JointModelSamples.R` to `R/JointModelQuantities.R`

Example of usage can be found in `design/tests/generated_quantities.R`

Things not implemented:

- Individual patient selection - This I think was much harder than I had originally anticipated as you need to correctly filter the entire data object and then also filter your  stan-samples to only contain the correct random effects.  Most of the run time I believe is spent on IO so it may be easier to just implement this by calculating the GQ for everyone and then filtering down to just the required patients (in Stan)

- In the next PR for the plotting functions I may try to move the call to `generateQuantites(samples)` inside of the autoplot function to spare the user having to do all this object juggling. However one advantage of having a dedicated user object is that it caches all the results for subsequent calls rather than having to keep reaching out to stan.

### DIFF
--- a/R/utilities.R
+++ b/R/utilities.R
@@ -250,7 +250,7 @@ expand_time_grid <- function(time_grid, max_time) {
         all(is.finite(time_grid)),
         msg = "`time_grid` needs to be finite, sorted, unique valued numeric vector"
     )
-    return(time_grid)
+    time_grid
 }
 
 


danielinteractive (commented):
thanks @gowerc looks very nice to me!

R/utilities.R
```diff
@@ -223,3 +223,62 @@ is_windows <- function() {
...
+    assert_that(
+        !any(is.na(time_grid)),
+        is.numeric(time_grid),
+        !is.null(time_grid),
+        !is.unsorted(time_grid),
+        !any(duplicated(time_grid)),
+        all(is.finite(time_grid)),
+        msg = "`time_grid` needs to be finite, sorted, unique valued numeric vector"
+    )
+    return(time_grid)
```
danielinteractive, line 253: minor style comment: best to avoid explicit `return()` statements if not needed, e.g. here just
```suggestion
    time_grid
```

danielinteractive (approved):
thanks Craig looks good!

PR #232, opened 2023-10-10
Enable Blank data objects

### LINKED ISSUE #192: Allow for blank data objects
Currently the `DataJoint` expects that both OS and LM data are available even if the model being fitted is only for 1 of them (e.g. no OS model is provided).  This is a bit clunky and would be good to allow DataJoint to work even if one of them isn't available.

gowerc:
Closes #192 

Still need to update the unit tests and documentation but I think this is feature complete.

Main change is the introduction of a `DataSubject` object which contains the essential per subject data e.g. `subject` / `arm` / `study`.   This in turn means `DataLongitudinal` and `DataSurvival` are now optional and can be omitted if the user is using a non-joint model (single model ?) 

The current api for a full joint model is:
```
DataJoint(
    subject = DataSubject(
        data = dat_os,
        subject = "pt",
        arm = "arm",
        study = "study"
    ),
    longitudinal = DataLongitudinal(
        data = dat_lm,
        formula = sld ~ time,
        threshold = 5
    )
    survival = DataSurvival(
        data = dat_os,
        formula = Surv(time, event) ~ cov_cat + cov_cont
    )
)
```

Notes on some internal design decisions:

- Users no longer specify the `subject` variable in `DataSurvival` nor `DataLongitudinal` instead it is assumed that the variable specified in `DataSubject` is the same variable in the other 2 objects (to save the user having to write it out 3 times)
- The internal method `suit_up`  is used to align the datasets, namely it casts `subject` in each dataset to factor with the same levels to ensure consistency; it also sorts the datasets by `subject` (and `time` in the case of `DataLongitudinal) again to ensure consistency
- Had to change the `as.list` method on the `Data*` objects to a custom `as_stan_list` method as I now need to pass arguments to them to inform them which variable to get subject information from. `as.list`'s default (for the case of NULL objects) method converts additional arguments to variables in the list where as I needed it to just return NULL

### DIFF
--- a/.lintr
+++ b/.lintr
@@ -1,3 +1,4 @@
+
 linters: linters_with_defaults(
     line_length_linter(120),
     object_name_linter = NULL,     # Stops complaints about functions not being in snake_case
@@ -5,4 +6,8 @@ linters: linters_with_defaults(
     cyclocomp_linter = NULL,
     indentation_linter = indentation_linter(4),
     object_length_linter = object_length_linter(40)
+    )
+exclusions: list(
+    "design/tests",
+    "./inst/WORDLIST"
     )
--- a/R/LinkRandomSlope.R
+++ b/R/LinkRandomSlope.R
@@ -23,7 +23,7 @@ NULL
 #'
 #' @export
 LinkRandomSlope <- function(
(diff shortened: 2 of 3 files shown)


danielinteractive (dismissed):
thanks @gowerc, looks good!

PR #234, opened 2023-10-12
Prior parameters supplied as data

### LINKED ISSUE #106: Convert priors to be input data to avoid unnecesary model compilation
Currently priors specified by the user get translated into direct stan code. That is to say they are hardcoded into the stan model itself meaning if a user changes a parameter on a prior the whole model needs to be re-compiled even if the distribution wasn't changed.  It would be good if we could pass prior parameters as data to avoid this. This might not be trivial however as it may be complicated to avoid namespace collisions of the prior input names.

gowerc:
Closes #106 

Is feature complete but still need to do docs + tests so currently WIP

### DIFF
--- a/R/DataJoint.R
+++ b/R/DataJoint.R
@@ -123,9 +123,9 @@ setValidity(
 #' Data Object -> `list`
 #'
 #' @param object (`DataSubject` or `DataLongitudinal` or `DataSurvival`) \cr
-#' data object to convert to a `list`
+#' data object to convert to a `list`.
 #' @param x (`DataSubject` or `DataLongitudinal` or `DataSurvival`) \cr
-#' data object to convert to a `list`
+#' data object to convert to a `list`.
 #' @param subject_var (`character`) \cr the name of the variable
 #' containing the subject identifier.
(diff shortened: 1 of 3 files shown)


R/DataJoint.R
```diff
@@ -117,7 +117,25 @@ setValidity(
...
-#' @rdname as_stan_list
+
+
+
+#' Data Object -> `list`
+#'
+#' @param object (`DataSubject` or `DataLongitudinal` or `DataSurvival`) \cr
+#' data object to convert to a `list`
+#' @param x (`DataSubject` or `DataLongitudinal` or `DataSurvival`) \cr
+#' data object to convert to a `list`
```
danielinteractive, line 128:
```suggestion
#' @param object (`DataSubject` or `DataLongitudinal` or `DataSurvival`) \cr
#' data object to convert to a `list`.
#' @param x (`DataSubject` or `DataLongitudinal` or `DataSurvival`) \cr
#' data object to convert to a `list`.
```
(just for consistency)

danielinteractive (approved):

PR #237, opened 2023-10-19
Feature/print quantities

### LINKED ISSUE #197: Create Print Methods
- [x] LongitudinalModel
- [x] Link
- [x] SurvivalModel
- [x] Joint Model
- [x] Parameter
- [x] Parameter List
- [x] Prior Objects
- [ ] JointSamples
- [ ] StanModule
- [ ] LongitudinalQuantities
- [ ] SurvivalQuantities
- [ ] Quantities
- [x] DataJoint
- [x] DataSubject
- [x] DataSurvival
- [x] DataLongitudinal

gowerc:
Closes #197 

Implements the final missing print methods for:

- Quantities
- LongitudinalQuantities
- SurvivalQuantities
- JointModelSamples
- StanModule

Apologies after looking through the documentation I realised I had used the legacy function `expect_snapshot_output` in a couple of places which I had to change to `expect_snapshot` but this had the side effect of changing some of the existing stored print snapshots as `expect_snapshot` also stores the code used to generate them. As far as I can tell though none of the actual output content has changed.

### DIFF
--- a/tests/testthat/test-LongitudinalQuantiles.R
+++ b/tests/testthat/test-LongitudinalQuantiles.R
@@ -80,7 +80,7 @@ test_that("autoplot.LongitudinalQuantities works as expected", {
 
 
 
-test_that("LongitudinalQuantities print method works as expected",{
+test_that("LongitudinalQuantities print method works as expected", {
 
     expect_snapshot({
         ptgroups <- c("pt_00011", "pt_00061", "pt_00001", "pt_00002")


tests/testthat/test-SurvivalQuantities.R
```diff
@@ -162,4 +162,25 @@ test_that("smoke test for summary(SurvivalQuantities) and autoplot(SurvivalQuant
     expect_true(inherits(p$layers[[2]]$geom, "GeomKm"))
     expect_true(inherits(p$layers[[3]]$geom, "GeomKmTicks"))
 
+
+    #########################
```
danielinteractive, line 166: Maybe nicer to put this into a separate `test_that()` call to avoid this comment dividers?

danielinteractive (approved):

tests/testthat/test-SurvivalQuantities.R
danielinteractive: Yeah in that case can be helpful to either prepare the object outside the test and then reuse in below tests. Or if it is fast to execute put I to a test helper function and the execute in each test

PR #241, opened 2023-11-02
Implements the Brier Score

### LINKED ISSUE #7: KM and scaled_brier
Implement functions `KM` (for Kaplan Meier curve) and  `scaled_brier` ( for scaled brier score estimation).

gowerc:
Closes #7

### DIFF
--- a/inst/WORDLIST
+++ b/inst/WORDLIST
@@ -106,3 +106,8 @@ harmonise
 JointModelSamples
 brierScore
 IPCW
+cdot
+le
+lt
+gt
+mathbb


R/SurvivalQuantities.R
```diff
@@ -356,3 +356,60 @@ setMethod(
...
     }
 )
+
+
+
+#' `brierScore`
+#'
+#' @description
+#' Derives the Brier Scores (using Inverse Probability of Censoring Weighting)
+#' for the Survival estimates as deatailed in Blanche et. al. (2015).
```
danielinteractive:
```suggestion
#' for the Survival estimates as detailed in Blanche et. al. (2015).
```

danielinteractive: as next step / new issue could consider using `Rdpack` (see https://geobosh.github.io/Rdpack/ and https://github.com/openpharma/mmrm for example) for references such as this one

R/brier_score.R
```diff
@@ -0,0 +1,241 @@
...
+        event_offset = event_offset,
+        maintain_cen_order = maintain_cen_order
+    )
+
+    # the following is a computational shortcut for diag(A %*% B)
+    # as we don't want to compute the off-diagonal entries of the
+    # matrix multiplication
+    x <- colSums(weight_mat * square_diff_mat)
+    names(x) <- t
+    return(x / length(times))
```
danielinteractive, line 163:
```suggestion
    x / length(times)
```
good to avoid explicit return where not needed

R/brier_score.R
```diff
@@ -0,0 +1,241 @@
...
+}
+
+
+#' @rdname brier_score
+bs_get_squared_dist <- function(t, times, events, pred_mat) {
+
```
danielinteractive, line 169: assertions here could make it more robust before going into the calculations below

R/brier_score.R
```diff
@@ -0,0 +1,241 @@
...
+#' @rdname brier_score
+bs_get_weights <- function(
+    t,
+    times,
+    events,
+    event_offset = TRUE,
+    maintain_cen_order = TRUE
+) {
+
```
danielinteractive, line 196: also here

tests/testthat/test-brierScore.R
```diff
@@ -0,0 +1,252 @@
+
+test_that("brierScore(SurvivalQuantities) returns same results as survreg", {
+    ### Setup Data
```
danielinteractive, line 3: maybe good to add here more explanation as comments why this should be the same (e.g. is it a special case of the Bayesian model etc.)

danielinteractive (commented):
Thank you @gowerc , looks very nice!

I wonder if it would be possible to add a methods details vignette where you describe the Brier Score calculations? That would make it easier for future developers before jumping into the code

danielinteractive (approved):

PR #249, opened 2023-12-15
Improve the GSF model implementation

### LINKED ISSUE #123: GSF model converges but has poor diagnostic properties

gowerc:
Closes #123 

- Extends the GSF model to have a `centered` argument so that users can toggle between the centered and non-centered parameterisations

- There wasn't a natural conversion between centered and non-centered for the phi parameter so I have converted that to be a beta distribution in both cases (it appears to be stable in this formulation regardless.

- Introduced a cron job that runs twice a month that runs the full joint GSF model as this unit test currently takes 30 minutes to run so didn't want to include it as part of the default run. This test checks that the model can recover known parameter values within the full model

- Fixed a bug where the sparse matrix representation of subject -> visit mapping for the longitudinal data was incorrectly using a full intercept column instead of correctly specifying the index's for patient 1 in the first column

- A handful of minor tweaks (added rounding to print messages, fixed some minor bugs that were blocking unit tests finishing)

- Converted the GSF simulation function to just use the centered parameterisation (there was literally no point in this function using the non-centered parameterisation, it just made the code more complicated). 

- Includes a directory of example models that build up the GSF model in stages (can be removed as not needed for the package but figured it might be useful if we need to experiment in the future). 

- Added an up-to-date outline of the GSF model specification to the stats-specification vignette

- For the non-centered parameterisation I converted the means to be on the log-scale so that the parameter estimates are consistent with the estimates in the centered parameterisation (as it is the log(mean) that you specify to the lognormal distribution).

### DIFF
--- a/vignettes/statistical-specification.Rmd
+++ b/vignettes/statistical-specification.Rmd
@@ -33,9 +33,6 @@ contain complete information for this package yet.
 
 
 
-
-
-
 # Longitudinal Model Specification
 
 
@@ -49,10 +46,10 @@ SLD_{ij} &=b_{i}
     \phi_i e^{-s_{i}t_{ij}}+
     (1-\phi_{i}) e^ {g_{i}t_{ij}}
 \right] \\ \\
-b_i &\sim \text{LogNormal}(\mu_{bl}, \omega_b) \\
-s_i &\sim \text{LogNormal}(\mu_{sk}, \omega_s) \\
-g_i &\sim \text{LogNormal}(\mu_{gk}, \omega_g) \\
-\phi &\sim \text{Beta}(a_{\phi k}, b_{\phi k})
+b_i &\sim \text{LogNormal}(\mu_{bl(i)}, \omega_b) \\
+s_i &\sim \text{LogNormal}(\mu_{sk(i)}, \omega_s) \\
+g_i &\sim \text{LogNormal}(\mu_{gk(i)}, \omega_g) \\
+\phi &\sim \text{Beta}(a_{\phi k(i)}, b_{\phi k(i)})
 \end{align*}
 $$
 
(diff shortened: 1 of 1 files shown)


R/LongitudinalGSF.R
```diff
@@ -26,47 +26,94 @@ NULL
...
+        Parameter(name = "lm_gsf_omega_kg", prior = omega_kg, size = 1),
+
+        Parameter(name = "lm_gsf_a_phi", prior = a_phi, size = "n_arms"),
+        Parameter(name = "lm_gsf_b_phi", prior = b_phi, size = "n_arms"),
+        Parameter(name = "lm_gsf_psi_phi", prior = psi_phi, size = "Nind"),
+
+        Parameter(name = "lm_gsf_sigma", prior = sigma, size = 1)
+    )
+
+    if (centered) {
```
danielinteractive, line 95: better `assert_flag` on `centered` first

R/LongitudinalGSF.R
```diff
@@ -26,47 +26,94 @@ NULL
...
+        parameters_extra <- list(
```
danielinteractive, line 96: better follow pattern `parameters_extra` <- if (...) ... else ...` so move the assignment outside

vignettes/statistical-specification.Rmd
```diff
@@ -31,6 +31,55 @@ Please note that this document is currently a work-in-progress and does not
...
+* $i$ is the subject index
+* $j$ is the visit index
+* $y_{ij}$ is the observed tumour measurements
+* $SLD_{ij}$ is the expected sum of longest diameter for subject $i$ at time point $j$
+* $t_{ij}$ is the time since first treatment for subject $i$ at visit $j$
+* $b_i$ is the subject baseline SLD measurement
+* $s_i$ is the subject kinetics shrinkage parameter
+* $g_i$ is the subject kinetics tumour growth parameter
+* $\phi_i$ is the subject proportion of cells affected by the treatment
+* $k$ is the treatment arm index for subject $i$
```
danielinteractive, line 70: hm ok but then better write $k(i)$ (or maybe $a(i)$ for arm?) and $l(i)$ (or maybe $s(i)$ for study?) otherwise the notation above is not easy to understand

vignettes/statistical-specification.Rmd
```diff
@@ -31,6 +31,55 @@ Please note that this document is currently a work-in-progress and does not
...
+* $l$ is the study index for subject $i$
+* $\mu_{xk}$ is the population mean for parameter $x$ in group $k$
+* $\omega_{x}$ is the population parameter for parameter $x$.
+* $\eta_{xi}$ is a random effects offset on parameter $x$ for subject $i$
```
danielinteractive, line 74: I would avoid using $x$ as a parameter because that is usually the data.
Also what is "group" here? or maybe you mean study or arm?

danielinteractive (commented):
Thanks @gowerc , lots of work here, please see few comments below

danielinteractive (approved):

PR #251, opened 2024-01-30
Add Stan + jmpost examples

### LINKED ISSUE #250: Add examples comparing the UI/UX of JMpost to Stan

gowerc:
Closes #250

Attempts to address @mercifr1's request to have a working example showing how jmpost code compares to raw Stan code. Note that jmpost has significantly worse performance due to the fact that we calculate the survival function by approximating the integral of the hazard function (as opposed to the raw Stan code here that just uses the closed form of the survival function as this is an OS only model.)

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -25,7 +25,7 @@ License: Apache License (>= 2)
 Encoding: UTF-8
 Language: en-US
 Roxygen: list(markdown = TRUE)
-RoxygenNote: 7.2.3
+RoxygenNote: 7.3.1
 Depends:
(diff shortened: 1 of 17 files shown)


danielinteractive (commented):
Thanks @gowerc , looks good, however I wonder if we could directly have this as a vignette in the package itself? As the `design` folder won't be visible for users, and is just for us as developers?

danielinteractive:
I think it could be helpful for Stan users to see the analogies. But ok can keep in `design` for now

mercifr1 (commented):
Thank you so much!

PR #252, opened 2024-01-31
Add missing prior distributions

### LINKED ISSUE #247: Add additional Prior distributions
For reference we currently have the following prior distributions implemented:

- normal
- weibull
- exponential
- cauchy
- beta
- lognormal
- gamma
- std_normal (N(0,1)) 

@danielinteractive / @mercifr1  - are there any that stand out to you as missing e.g. common distributions that we should add in? Skimming the Stan documentation the following caught my eye that I wasn't sure if we would need or not:

- uniform
- gumbel
- student-t
- logistic
- skew-normal (?)
- Chi-square
- inverse-gamma
- Pareto
- Wishart (though multivariate distributions might require some rework of our current framework)

gowerc:
Closes #247

(well everything is implemented except the log-student-t pending conversation in the issue)

Summary of changes:

- Added uniform distribution
- Added logistic distribution
- Added loglogistic distribution
- Added inv_gamma distibution
- Added student-t distribution
- Fixed bug with `merge` method for `StanModule` that was causing test failures due to adding in empty components rather than stripping them off
- Added missing `@export` tag which the latest version of Roxygen now complains about and was causing documentation failure


R/StanModule.R
```diff
@@ -144,11 +144,13 @@ setMethod(
         stan_fragments <- lapply(
             names(stan_blocks),
             \(par) {
-                new_string <- c(slot(x, par), slot(y, par))
-                if (all(new_string == "")) {
-                    return("")
+                if (all(slot(y, par) == "")) {
```
danielinteractive, line 147: do you already have some unit tests for this specific thing? would it be worth to factor this into an explicit helper function outside of the `lapply()` to be able to test it better?

danielinteractive (commented):
nice work, thanks a lot @gowerc !

danielinteractive (dismissed):
Thanks, that is good (separate helper might still be a bit better but it's ok like this)

danielinteractive (approved):

PR #254, opened 2024-02-02
Different initial values for each mcmc chain

### LINKED ISSUE #4: Different initial values for each mcmc chain
With the current implementation, the initial values are part of the stanAll object. Since there is only one slot for the initial values, the user is unable to define different initial values for every mcmc run. It would be better if the user is able to define different values.

gowerc:
Closes #4

This ended up being quite a bit more complicated than expected, mostly because the init code was kind of scattered all throughout the program.

I think in particular I would appreciate a double checking of the sample functions as the parameterisations differ in a few cases between Stan and R soo need to double check I'm sampling them correctly.

### DIFF
--- a/R/Prior.R
+++ b/R/Prior.R
@@ -562,7 +562,8 @@ prior_invgamma <- function(alpha, beta) {
 #'
 #' R's sampling functions only produce the standard Student-t distribution so in order
 #' to match Stan's implementation we multiply by the scale parameter and add the location
-#' parameter. See https://stats.stackexchange.com/a/623611 for details
+#' parameter. See this \href{https://stats.stackexchange.com/a/623611}{Stack Overflow} post
+#' for details
 #'
 #' @name Local_Sample
 #' @keywords internal
--- a/man/Local_Sample.Rd
+++ b/man/Local_Sample.Rd
@@ -80,7 +80,8 @@ by sampling sampling from the Gamma distribution and then taking the reciprocal.
 
 R's sampling functions only produce the standard Student-t distribution so in order
 to match Stan's implementation we multiply by the scale parameter and add the location
-parameter. See https://stats.stackexchange.com/a/623611 for details
+parameter. See this \href{https://stats.stackexchange.com/a/623611}{Stack Overflow} post
+for details
 }
 }
 \keyword{internal}


R/JointModel.R
```diff
@@ -150,21 +152,36 @@ sampleStanModel.JointModel <- function(object, data, ...) {
...
-    if (!"init" %in% names(args)) {
-        values_initial <- initialValues(object)
-        values_sizes <- size(object@parameters)
-        values_sizes_complete <- replace_with_lookup(values_sizes, args[["data"]])
-        values_initial_expanded <- expand_initial_values(values_initial, values_sizes_complete)
-        args[["init"]] <- function() values_initial_expanded
+    args[["chains"]] <- if ("chains" %in% names(args)) {
+        args[["chains"]]
+    } else {
+        4
```
danielinteractive, line 164: where is this magic number coming from? could we introduce a package option for this instead?

R/LongitudinalGSF.R
```diff
@@ -87,17 +73,17 @@ LongitudinalGSF <- function(
 
         Parameter(name = "lm_gsf_a_phi", prior = a_phi, size = "n_arms"),
         Parameter(name = "lm_gsf_b_phi", prior = b_phi, size = "n_arms"),
-        Parameter(name = "lm_gsf_psi_phi", prior = psi_phi, size = "Nind"),
+        Parameter(name = "lm_gsf_psi_phi", prior = prior_none(), size = "Nind"),
```
danielinteractive, line 76: how would the user change the prior on this parameter now that we don't have it in the arguments anymore?

R/Parameter.R
```diff
@@ -102,6 +102,7 @@ as_stan_list.Parameter <- function(object, ...) {
 #'
 #' @param x (`Paramater`) \cr A model parameter
 #' @param object (`Paramater`) \cr A model parameter
+#' @param ... Not Used.
```
danielinteractive, line 105:
```suggestion
#' @param ... Not used.
```

R/ParameterList.R
```diff
@@ -132,6 +132,7 @@ as.list.ParameterList <- function(x, ...) {
 #' Getter functions for the slots of a [`ParameterList`] object
 #' @inheritParams ParameterList-Shared
 #' @family ParameterList
+#' @param nchains (`integer`) \cr the number of chains.
```
danielinteractive, line 135: or `n_chains`? just a question, best to go with what fits in naming pattern used so far

R/ParameterList.R
```diff
@@ -145,11 +146,17 @@ names.ParameterList <- function(x) {
...
-initialValues.ParameterList <- function(object) {
-    vals <- lapply(object@parameters, initialValues)
-    name <- vapply(object@parameters, names, character(1))
-    names(vals) <- name
-    return(vals)
+initialValues.ParameterList <- function(object, nchains, ...) {
+    x <- lapply(
+        seq_len(nchains),
+        \(i) {
+            vals <- lapply(object@parameters, initialValues)
```
danielinteractive, line 153: possible to use `vapply` here too or not?

danielinteractive: if not maybe add a comment

R/ParameterList.R
```diff
@@ -145,11 +146,17 @@ names.ParameterList <- function(x) {
 
 #' @describeIn ParameterList-Getter-Methods The parameter-list's parameter initial values
 #' @export
-initialValues.ParameterList <- function(object) {
-    vals <- lapply(object@parameters, initialValues)
-    name <- vapply(object@parameters, names, character(1))
-    names(vals) <- name
-    return(vals)
+initialValues.ParameterList <- function(object, nchains, ...) {
+    x <- lapply(
```
danielinteractive, line 150: can remove `x` assignment and directly have the `lapply()` as last value in this method

R/Prior.R
```diff
@@ -324,17 +338,17 @@ prior_beta <- function(a, b, init = a / (a + b)) {
...
 #' @export
-prior_none <- function(init = 0.00001) {
+prior_none <- function() {
     Prior(
         parameters = list(),
         display = "<None>",
         repr_model = "",
         repr_data = "",
-        init = init,
+        sample = \(n) local_runif(n, -4, 4),
```
danielinteractive, line 350: also here wondering if those magic numbers would better live in a package option

R/Prior.R
```diff
@@ -359,7 +372,8 @@ prior_uniform <- function(alpha, beta, init = 0.5 * (alpha + beta)) {
             "real prior_alpha_{name};",
             "real prior_beta_{name};"
         ),
-        init = init,
+        init = 0.5 * (alpha + beta),
+        sample = \(n) local_runif(n, alpha, beta),
         validation = list(
```
danielinteractive, line 377: do we check somewhere that alpha is smaller than beta?

R/Prior.R
```diff
@@ -479,10 +495,73 @@ prior_invgamma <- function(alpha, beta, init = beta / (alpha - 1)) {
...
+#' @rdname Local_Sample
+local_rlnorm <- \(...) rlnorm(...)
+
+#' @rdname Local_Sample
+local_rbeta <- \(...) rbeta(...)
+
+#' @rdname Local_Sample
+local_runif <- \(...) runif(...)
+
+#' @rdname Local_Sample
```
danielinteractive, line 550: I guess for these functions where there is some computation behind it, it would be good to document in a developer vignette where the formulas are coming from.

vignettes/model_fitting.Rmd
```diff
@@ -373,3 +370,45 @@ sq <- SurvivalQuantities(
 )
 brierScore(sq)
 ```
+
+
+## Initial Values
+
+By default `jmpost` will set the initial values for all parameters to be a random value
+drawn from the prior distribution that has been shrunk towards the
+mean of said distribution e.g. for a `prior_normal(4, 2)`
```
danielinteractive, line 379: would be good to be able to control the amount of shrinkage with a package option

danielinteractive (commented):
Thanks @gowerc ! nice work

R/Prior.R
danielinteractive: ah ok, I just think it would be more elegant to capture this in a package option, that defaults to `c(-4, 4)` but in principle the user could change it

R/JointModel.R
danielinteractive: yeah but instead of `4` we could read this from our package option? which defaults to 4?

danielinteractive: and if user change does not make sense - then let's have a constant defined and documented and use the constant here

danielinteractive (approved):

PR #259, opened 2024-02-14
Rework link functions

### LINKED ISSUE #238: Rework Link functions
In hindsight I think the way I have implemented the links is a bit short sighted.  For example there is no reason why the "identity" / "dSLD" links can't be used for the random slope model.  Likewise the Random Slope model would also benefit from the ability to stack multiple links like the GSF model. 

I imagine this would need to be reworked as a set of generic functions on the Longitudinal models where each longitudinal model can implement their version of a "identity".  I also image the link stacking needs to be generalised though this might be problematic to implement as it involves using jinja syntax within the longitudinal models which, whilst not too hard for us to implement, makes it quite a bit harder for others to then extend with their own model variants. 

My gut feeling is we can ignore this until after the initial internal release however I would argue it is likely a high priority item after said release. Likewise it potentially conflicts with #170 as that ticket requires us to add a bunch more link code which we would then need to revert.

gowerc:
Closes #238 

I think I've finished the core functionality but still need to re-do the docs and unit tests. Was just hoping to share to see if there was any initial concerns / feedback.

Theres a lot going on in this PR so the below attempts to highlight the core components + how it works:

In terms of the UI the end users would now call the modelling functions like:

```
jm <- JointModel(
    longitudinal = LongitudinalRandomSlope(),
    survival = SurvivalExponential(),
    link = link_dsld()
)
```

Or if they have multiple links:
```
jm <- JointModel(
    longitudinal = LongitudinalRandomSlope(),
    survival = SurvivalExponential(),
    link = Link(
        link_dsld(),
        link_identity(prior = prior_normal(0, 0.02))
    )
)
```

Going one layer down, the workhorse object is the `LinkComponent`.  The idea is these would be exported so users can define their own link code if needed:

```
link_bob <- function(prior = prior_normal(0, 2)) {
    LinkComponent(
        key = "link_bob",
        stan = StanModule("<some stan code here>"),
        parameter = ParameterList(Parameter(name = "link_bob", prior = prior, size = 1))
    )
}
```

In terms of the required stan code, the current interface for creating your own link is a stan function of the signature:
```
matrix <key>_contrib(
        matrix time,
        matrix link_function_inputs
)
```
Where `link_function_inputs` is an arbitrary matrix that is defined by the longitudinal model. For example for the GSF model it is defined as:
```
    matrix[Nind, 4] link_function_inputs;
    pars_lm[,1] = lm_gsf_psi_bsld;
    pars_lm[,2] = lm_gsf_psi_ks;
    pars_lm[,3] = lm_gsf_psi_kg;
    pars_lm[,4] = lm_gsf_psi_phi;
``` 
Technically it is possible for users to extend the `link_function_inputs` matrix from the default though the process of doing so is a little involved. That is, under the hood the longitudinal models define this Stan code by implementing the `enableLink()` method, so users can subclass the model and then implement their own `enableLink()` method to override the default. For reference this is the current enableLink for the RandomSlope model:
```
enableLink.LongitudinalRandomSlope <- function(object, ...) {
    object@stan <- merge(
        object@stan,
        StanModule("lm-random-slope/link.stan")
    )
    object
}
```

In terms of our implementation of the common links (e.g. dsld/ttg)  we allow the `stan` argument to `LinkComponent `to be a method which will be called against the Longitudinal model object allowing the models to provide their own implementations of each link. For example the actual `link_dsld()` function is:

```
#' @export
link_dsld <- function(prior = prior_normal(0, 2)) {
    LinkComponent(
        key = "link_dsld",
        stan = linkDSLD,
        parameter = ParameterList(Parameter(name = "link_dsld", prior = prior, size = 1))
    )
}
```

And then we implement model specific methods to provide individual model implementations e.g.
```
linkDSLD.LongitudinalRandomSlope <- function(object, ...) {
    StanModule("lm-random-slope/link_dsld.stan")
}
```

And finally as all models subclass from the base `LongitudinalModel` class I've also provided default implementations to throw meaningful errors if a model doesn't have a method defined e.g.

```
linkDSLD.LongitudinalModel <- function(object, ...) {
    stop(
        sprintf(
            "DSLD link is not available for the %s model",
            object@name
        )
    )
}
```

### DIFF
--- a/vignettes/extending-jmpost.Rmd
+++ b/vignettes/extending-jmpost.Rmd
@@ -166,7 +166,7 @@ Under the hood these link-family functions work by defining a `LinkComponent` ob
 that specifies a method that will return a `StanModule` object instead of specifying
 the `StanModule` object directly. This method will then be called by the `JointModel` constructor
 with the `LongitudinalModel` object as its first argument. This way each longitudinal model is 
-able to specify their own implementation for the given link family. The following is rough outline
+able to specify their own implementation for the given link family. The following is a rough outline
 for how this works for the `link_dsld()` function and the `LongitudinalGSF` model:
 
 ```R
@@ -183,8 +183,8 @@ linkDSLD.LongitudinalGSF <- function(object, ...) {
 }
 ```
 
-This is to say that if you wish to overwrite a models implementation for existing link family or
-provide an implementaiton for a new model then you can do so by defining/overwriting the relevant
+This is to say that if you wish to overwrite a model implementation for an existing link family or
+provide an implementation for a new model then you can do so by defining/overwriting the relevant
 methods. For clarity the underlying methods are `linkDSLD`, `linkIdentity` and `linkTTG`
 respectively.
 


danielinteractive:
Just read through the description, and looks very good to me @gowerc !

R/LongitudinalModel.R
```diff
@@ -51,3 +51,34 @@ as_print_string.LongitudinalModel <- function(object, ...) {
     )
     return(string)
 }
+
+
+#' @export
+linkTTG.LongitudinalModel <- function(object, ...) {
```
danielinteractive, line 57: this works well for the packaged links, I wonder if this could be positioned "higher" in the methods tree, such that when the user / externally defined link is called without the method implementation it would also give a sensible error message?

vignettes/extending-jmpost.Rmd
```diff
@@ -28,6 +28,151 @@ library(jmpost)
 Please note that this document is currently a work-in-progress and does not
 contain complete information for this package yet. 
 
+
+## Custom Link Functions
+
+Users can defined custom link functions in several ways based upon the level of customisation
```
danielinteractive, line 34:
```suggestion
Users can define custom link functions in several ways based upon the level of customisation
```

vignettes/extending-jmpost.Rmd
```diff
@@ -28,6 +28,151 @@ library(jmpost)
...
+    link_function_inputs[,1] = lm_gsf_psi_bsld;
+    link_function_inputs[,2] = lm_gsf_psi_ks;
+    link_function_inputs[,3] = lm_gsf_psi_kg;
+    link_function_inputs[,4] = lm_gsf_psi_phi;
+}")
+    object@stan <- merge(
+        object@stan,
+        stan
+    )
+    object
```
danielinteractive, line 75: side comment (not for this PR): it might be nice to have an `update(object, stan)` method that does this, in a bit more readable and concise way

vignettes/extending-jmpost.Rmd
```diff
@@ -28,6 +28,151 @@ library(jmpost)
...
+```
+
+Note that there are a few families of link functions that are common across all models for example
+identity, dSLD, TTG. Users can access these by simply using the inbuilt `link_identity()`,
+`link_dsld()` and `link_ttg()` functions respectively.
+These functions are responsible for loading the correct `LinkComponent` to implement that link for a
+particular model.
+
+If you wish to provide implementations of these
+for new models then you need to provide method implementations for your model to `linkDSLD()`,
```
danielinteractive, line 154: maybe worth commenting on the difference of the snake case functions above and these camel case versions here

danielinteractive: or is that something to fix here?

danielinteractive (commented):
thanks @gowerc , looks nice, see a few minor comments below

R/LongitudinalModel.R
danielinteractive: Yes, exactly - what happens if the user defines a `link_xyz()` function and then there is no method for object `GSF`? Is that caught? (I guess not)

vignettes/extending-jmpost.Rmd
```diff
@@ -147,32 +147,48 @@ Link(
...
+    link_ttg()
+  )
+) 
+```
+
+Under the hood these link-family functions work by defining a `LinkComponent` object
+that specifies a method that will return a `StanModule` object instead of specifying
+the `StanModule` object directly. This method will then be called by the `JointModel` constructor
+with the `LongitudinalModel` object as its first argument. This way each longitudinal model is 
+able to specify their own implementation for the given link family. The following is rough outline
```
danielinteractive, line 169:
```suggestion
able to specify their own implementation for the given link family. The following is a rough outline
```

vignettes/extending-jmpost.Rmd
```diff
@@ -147,32 +147,48 @@ Link(
...
-        matrix[nrows, ncols] rnd_slope_mat = rep_matrix(lm_rs_ind_rnd_slope, ncols);
-        return rnd_slope_mat;
-    }
-}")
+  StanModule("<stan code for GSF dSLD link>")
 }
 ```
 
+This is to say that if you wish to overwrite a models implementation for existing link family or
+provide an implementaiton for a new model then you can do so by defining/overwriting the relevant
```
danielinteractive, line 187:
```suggestion
provide an implementation for a new model then you can do so by defining/overwriting the relevant
```

vignettes/extending-jmpost.Rmd
```diff
@@ -147,32 +147,48 @@ Link(
...
-        vector[nrows] lm_rs_ind_rnd_slope = link_function_inputs[,2];
-        matrix[nrows, ncols] rnd_slope_mat = rep_matrix(lm_rs_ind_rnd_slope, ncols);
-        return rnd_slope_mat;
-    }
-}")
+  StanModule("<stan code for GSF dSLD link>")
 }
 ```
 
+This is to say that if you wish to overwrite a models implementation for existing link family or
```
danielinteractive, line 186:
```suggestion
This is to say that if you wish to overwrite a model implementation for an existing link family or
```

danielinteractive (approved):

PR #260, opened 2024-02-19
Change LICENSE.md to plain text LICENSE

cicdguy:
To prevent linter complaints and to keep the license format consistent across Roche packages.


danielinteractive:
Can we maybe just remove this? CRAN e.g. will ask later anyway when we don't have modifications in the file

PR #262, opened 2024-02-21
More examples of AIC / BIC / loo

gowerc:
This PR adds more examples of using AIC / BIC / LOO as well as an log-logistic example. 

I don't think there are any meaningful changes to the package here so more just an FYI.

Can't be merged until #259  has been merged as its built on top of that branch.


danielinteractive (approved):
sounds good, thanks Craig

PR #265, opened 2024-02-23
Add Stein-Fojo Model

### LINKED ISSUE #171: Add simplified SF model (that is without the G)

gowerc:
Closes #171


danielinteractive (commented):
Thanks @gowerc , any chance we can exploit the relationship between SF and GSF? E.g. SF being a special case of GSF?

danielinteractive:
Ok for now then. I guess as a future idea one could in general enable "constant" priors as it seems brms also has which is then dynamically generating corresponding transformed parameters code. This could be useful in general and as a side product make GSF with constant prior on the mixture weight give the SF model.

PR #272, opened 2024-02-28
Added user option to adjust GQ function

### LINKED ISSUE #180: Add flexibility to Gaussian Quadrature parameter
Currently the Gaussian Quadrature process for estimating the survival distribution is hard coded to have `n=15`  I don't know much about the theory here but is this some universally agreed amount that is "good enough" or should this really be a user defined option ?

gowerc:
Closes #180

### DIFF
--- a/tests/testthat/test-options.R
+++ b/tests/testthat/test-options.R
@@ -25,7 +25,7 @@ test_that("Can alter Gaussian Quadrature arguments", {
 
 
     ## Test modified values
-    options( "jmpost.gauss_quad_n" = 20)
+    options("jmpost.gauss_quad_n" = 20)
     df <- DataSurvival(
         data = x,
         formula = Surv(vtime, vevent) ~ vcov1 * vcov2


R/DataSurvival.R
```diff
@@ -125,7 +125,10 @@ as_stan_list.DataSurvival <- function(object, ...) {
     rownames(design_mat) <- NULL
 
     # Parameters for efficient integration of hazard function -> survival function
-    gh_parameters <- statmod::gauss.quad(n = 15, kind = "legendre")
+    gh_parameters <- statmod::gauss.quad(
+        n = getOption("jmpost.gauss_quad_n"),
+        kind = getOption("jmpost.gauss_quad_kind")
```
danielinteractive, line 130: I don't think we can use something else than `legendre`, because otherwise the integration boundaries change and we would need to adjust computations on the Stan side. Just the `n` option is sufficient

PR #278, opened 2024-03-01
Simulation function supports multiple studies

### LINKED ISSUE #113: Update simulation function to support multiple studies
Simulation function currently only supports 1 study (but multiple arms within that study).  Need to look into updating this to  support the creation of multiple studies.

This is of particular importance for testing the D2 model as the original implementation had hierarchical parameters per study (not just arm)

gowerc:
Closes #113

Apologies lots of files ended up being edited but I promise its not as bad as it looks!  Its more just a symptom that I changed the api for `simulate_joint_data()` and that function is unfortunately used everywhere.

I opted to making it a bit more explicit so that users now have to explicitly define groups and state which arm/study those groups belong to e.g. 

```
    design = list(
        SimGroup(n = 50, study = "Study-1", arm = "Arm-A"),
        SimGroup(n = 50, study = "Study-1", arm = "Arm-B")
    )
```

The vast majority of the file changes are me then changing all the calls to meet this pattern.

Given this change I decided to update the RandomSlope model to have a separate intercept term per "study". The remainder of the file changes (mostly the Stan file changes) are simply just supporting this.

### DIFF
--- a/R/simulations.R
+++ b/R/simulations.R
@@ -43,8 +43,8 @@ get_timepoints <- function(x) {
     "SimGroup",
     slots = c(
         n = "numeric",
-        study = "character",
-        arm = "character"
+        arm = "character",
+        study = "character"
     )
 )
 
--- a/tests/testthat/test-simulations.R
+++ b/tests/testthat/test-simulations.R
@@ -252,3 +252,25 @@ test_that("simulate_joint_data correctly generates an intercept per study", {
     z_score <- (ests - c(50, 70, 10)) / ests_se
     expect_true(all(abs(z_score) < qnorm(0.99)))
 })
+
+
+test_that("SimGroup() works as expected", {
+    sim_group <- SimGroup(100, "Arm-A", "Study-X")
+    expect_s4_class(sim_group, "SimGroup")
+    expect_equal(sim_group@n, 100)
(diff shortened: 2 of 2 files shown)


R/simulations.R
```diff
@@ -23,10 +23,64 @@ get_timepoints <- function(x) {
...
+#' @examples
+#' SimGroup(n = 50, arm = "Arm-A", study = "Study-1")
+#' @name SimGroup-class
+#' @exportClass SimGroup
+.SimGroup <- setClass(
+    "SimGroup",
+    slots = c(
+        n = "numeric",
+        study = "character",
+        arm = "character"
```
danielinteractive, line 47: swap these two lines for consistency?

tests/testthat/test-simulations.R
```diff
@@ -204,3 +216,39 @@ test_that("sim_os_weibull creates a dataset with the correct parameter", {
     z_score <- (log_scale_obs - log_scale_real) / log_scale_se
     expect_true(abs(z_score) < qnorm(0.9))
 })
+
```
danielinteractive, line 219: would be good to have a few simple tests for the `SimGroup` class itself maybe

danielinteractive (approved):

PR #282, opened 2024-03-08
Support Negative Time / Re-work Simulation function

### LINKED ISSUE #215: Support pre-dose / negative time
Original GSF model in @mercifr1's paper had provisions for negative time / time before treatment. It might be worth us adding support for this. Currently we only support time from first dose.

gowerc:
Closes #215 

This is still a WIP

This ended up being more complicated than expected. Well its just that the current simulation setup could support negative times it was just very clunky and non-intuitive so I took this as an opportunity to overhaul the simulation functions and bring them to parity with the regular JointModel functions (e.g. the api is a lot closer now with key terms being the same with simply a `Sim` prefix). 

Am still updating the test files at the moment as unfortunately the simulation functions are used in a lot of places. Will provide a more broader explanation once I've finished coding.

### DIFF
--- a/R/SimSurvival.R
+++ b/R/SimSurvival.R
@@ -10,6 +10,7 @@
 #' @param beta_cont (`number`)\cr the continuous covariate coefficient.
 #' @param beta_cat (`numeric`)\cr the categorical covariate coefficients.
 #' @param loghazard (`function`)\cr the log hazard function.
+#' @param name (`character`)\cr the name of the object.
 #' @param ... Not Used.
 #'
 #' @section Hazard Evaluation:
@@ -41,6 +42,7 @@ NULL
 #' @slot beta_cont (`numeric`)\cr See arguments.
 #' @slot beta_cat (`numeric`)\cr See arguments.
 #' @slot loghazard (`function`)\cr See arguments.
+#' @slot name (`character`)\cr See arguments.
 #'
 #' @family SimSurvival
 #' @exportClass SimSurvival
--- a/man/SimSurvival-Shared.Rd
+++ b/man/SimSurvival-Shared.Rd
@@ -16,6 +16,8 @@
 
 \item{loghazard}{(\code{function})\cr the log hazard function.}
 
+\item{name}{(\code{character})\cr the name of the object.}
+
 \item{...}{Not Used.}
 }
 \description{
--- a/man/SimSurvival-class.Rd
+++ b/man/SimSurvival-class.Rd
@@ -29,6 +29,8 @@ SimSurvival(
 \item{beta_cat}{(\code{numeric})\cr the categorical covariate coefficients.}
 
 \item{loghazard}{(\code{function})\cr the log hazard function.}
+
+\item{name}{(\code{character})\cr the name of the object.}
 }
 \description{
 Abstract Simulation Class for Survival Data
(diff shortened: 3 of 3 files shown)


R/SimLongitudinalGSF.R
```diff
@@ -0,0 +1,188 @@
+
+#' Simulate Longitudinal Data from a  GSF Model
```
danielinteractive, line 2:
```suggestion
#' Simulate Longitudinal Data from a GSF Model
```

R/SimLongitudinalGSF.R
```diff
@@ -0,0 +1,188 @@
...
+#'
+#' @param times (`numeric`)\cr the times to generate observations at.
```
danielinteractive, line 4: I think would be worth to consider `roxytypes`, now that it is on CRAN :) (separate PR of course)

R/SimLongitudinalGSF.R
```diff
@@ -0,0 +1,188 @@
...
+            length(object@b_phi)
+        )
+        if (length(unique(par_lengths)) != 1) {
+            return("The parameters `mu_s`, `mu_g`, `a_phi`, and `b_phi` must have the same length.")
+        }
+        if (length(object@sigma) != 1) {
+            return("The parameter `sigma` must have length 1.")
+        }
+        if (length(c(object@omega_b, object@omega_s, object@omega_g)) != 3) {
+            return("The parameters `omega_b`, `omega_s`, and `omega_g` must be length 1.")
```
danielinteractive, line 103: this is not very precise. You could have two of them with length 0 and one with length 3

R/SimLongitudinalGSF.R
```diff
@@ -0,0 +1,188 @@
...
+        }
+        if (length(c(object@link_dsld, object@link_ttg, object@link_identity)) != 3) {
```
danielinteractive, line 105: similar here and maybe elsewhere

R/SimLongitudinalGSF.R
```diff
@@ -0,0 +1,188 @@
...
+#'
+#' @keywords internal
+gsf_sld <- function(time, b, s, g, phi) {
+    phi <- dplyr::if_else(time >= 0, phi, 0)
+    b * (phi * exp(-s * time) + (1 - phi) * exp(g * time))
+}
+
+
+#' @rdname gsf_sld
+gsf_ttg <- function(time, b, s, g, phi) {
```
danielinteractive, line 175: these kinds of formulas would be good to document in math in the relevant vignette (or precursor design doc) - maybe you have that already though, just thinking out loud here

inst/stan/base/functions.stan
```diff
@@ -81,5 +81,24 @@ functions {
         }
         return ret;
     }
+
+    // Replaces the values of `y` with `replacement` if the coressponding value
```
danielinteractive, line 85:
```suggestion
    // Replaces the values of `y` with `replacement` if the corresponding value
```

danielinteractive (commented):
Thanks @gowerc , please see few comments inside, if you have something specific you would like me to have a careful look at please let me know :) (a bit too large for a detailed review indeed :))

danielinteractive (commented):
Thanks @gowerc , stan code changes look good to me. It is also important to have a few tests regarding negative and zero values in place, but probably you already have those I guess.

Issue #288, opened 2024-03-22
Add flexibility to generated quantities


gowerc:
The current implementation of the generated quantities functions allows the users to provide a list of patients and a time grid as 2 independent arguments.  The functions then predict the values for all patients at the given time grid.  

Unfortunately this doesn't quite meet two common use cases which would be:

- Predict just at the observed time points for each patient (which is different per patient)
- Predict at say X evenly spaced intervals between 0 and the patients event time

I think we potentially need 3 changes

- [x] Generalise the looping mechanism in the generated quantities to avoid each model having to re-implement the same logic
- [ ] Allow user arguments to generated quantities to specify different time points per subject
- [ ] Have helper objects / functions to setup the per patient grids as specified above

PR #289, opened 2024-03-22
Re-factored the stan code for longitudinal generated quantities

gowerc:
Addresses part 1 of #288 

In this PR I factor out the looping code into the base longitudinal model stan file.  This means each individual model doesn't have to concern itself about how patients are subset meaning less chance of issues in the future when we update this mechanism.

We then introduce a new interface that the longitudinal models have to conform to which is that they must create a function with signature:
```
row_vector lm_predict_individual_patient(vector time, row_vector long_gq_parameters)
```

They also must define a matrix `long_gq_parameters` which has 1 row per patient and as many columns as required for the model in question. As an example for the SF model this is defined as:
```
    matrix[Nind, 3] long_gq_parameters;
    long_gq_parameters[, 1] = lm_gsf_psi_bsld;
    long_gq_parameters[, 2] = lm_gsf_psi_ks;
    long_gq_parameters[, 3] = lm_gsf_psi_kg;
```

I also added a new unit test to show that patient subsetting works as expected and that the correct values are reconstructed per each sample per patient.

### DIFF
--- a/R/StanModule.R
+++ b/R/StanModule.R
@@ -262,16 +262,17 @@ is_file <- function(filename = NULL) {
 #' @param string Character, either the absolute path of a stan file, or the name of the stan
 #' file in the package directory or the stan code as a string.
 read_stan <- function(string) {
-    string <- paste0(string, collapse = "\n")
     local_inst_file <- file.path("inst", "stan", string)
     system_file <- system.file(file.path("stan", string), package = "jmpost")
     local_file <- string
     files <- c(local_file, local_inst_file, system_file)
     for (fi in files) {
         if (is_file(fi)) {
-            return(readLines(fi))
+            string <- readLines(fi)
+            break
         }
     }
+    string <- paste0(string, collapse = "\n")
     return(string)
 }
 


R/LongitudinalModel.R
```diff
@@ -28,13 +28,19 @@ LongitudinalModel <- function(
     ...
 ) {
 
-    base_long <- StanModule(
-        x = "base/longitudinal.stan"
+    base_stan <- paste0(
+        read_stan("base/longitudinal.stan"),
+        collapse = "\n"
```
danielinteractive, line 33: side question: is this pasting/collapsing really needed? I would have thought this kind of stuff is handled in `read_stan` & co.?

tests/testthat/test-LongitudinalQuantiles.R
```diff
@@ -124,3 +124,44 @@ test_that("LongitudinalQuantities can recover known results", {
 
     expect_true(all(dat_all$correl > 0.99))
 })
+
+
+test_that("LongitudinalQuantities correctly subsets patients and rebuilds correct value for each smaple", {
```
danielinteractive, line 129:
```suggestion
test_that("LongitudinalQuantities correctly subsets patients and rebuilds correct value for each sample", {
```

danielinteractive (dismissed):

PR #294, opened 2024-03-27
Rework of Generated Quantities

### LINKED ISSUE #288: Add flexibility to generated quantities

gowerc:
Closes #288 

In the current code you would specify subjects and time points independently and the timepoints would be generated for all subjects. Unfortunately this was too limiting as a request from Francois was to generate samples at the observed timepoints which are potentially different for every subject. This would lead to an excessive amount of unnecessary computation.

This change allows greater flexibility in how times and subjects are specified. The main change is in the UI which introduces a new set of `Grid()` objects which define what timepoints / subjects need to be sampled.  Currently I have introduced:

- `GridObserved(subjects)` - Generates a sample at each of the patients observation time points
- `GridFixed(subjects, times)` - Generates a sample at each time per patient (same to what we currently have)
- `GridGrouped(groups, times)` - Same as `GridFixed()` but allows for patients to be averaged across groups

In the future I will probs add `GridManual()` where you can specify individual times for specific subjects but will do that in a later PR. 

To support these changes there was quite a large change of the quantity internals as the equal number of times was a very hard baked assumption in the code.

### DIFF
--- a/R/GridFixed.R
+++ b/R/GridFixed.R
@@ -28,7 +28,7 @@ as.QuantityGenerator.GridFixed <- function(object, data, ...) {
     assert_class(data, "DataJoint")
     data_list <- as.list(data)
     subjects <- unlist(as.list(object, data = data), use.names = FALSE)
-    time_grid <- expand_time_grid(object@times, max(data_list[["Tobs"]]))
+    time_grid <- expand_time_grid(object@times, max(data_list[["tumour_time"]]))
 
     pt_times <- expand.grid(
         pt = subjects,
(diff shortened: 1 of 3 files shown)


danielinteractive (approved):
Thanks @gowerc , looks good in terms of UI and browsing through no red flags

danielinteractive (approved):

PR #299, opened 2024-04-03
Test commit from branch Francoistest1

mercifr1:
Test of branch


R/brier_score.R
```diff
@@ -1,4 +1,5 @@
 
+#' Test
```
mercifr1, line 2:
```suggestion
#' Test 3 would be better
```

Issue #296, opened 2024-03-28
Add support for generating quantities for the population level


gowerc:
Currently we only support generated quantities for individual subjects who exist within the dataset. Francois has asked that we also have functionality to produce estimates for the population level (e.g. a hypothetical patient with 0 random effects).

danielinteractive:
Thanks @gowerc , yeah agree we have to specify this in detail first before proceeding to the implementation. A general question here will be how much and what should be easily available by options/methods and what the user need to do themselves based on samples of all model parameters.

mercifr1:
Hi @gowerc 
The phi parameter is indeed a special case, as - unlike with the other parameters - it is not parameterized hierarchically. Actually, you could probably parameterize it hierarchically, and it would solve this issue. For instance, you could consider:
`
parameters{
real mean_mu_phi;
...
real<lower=0, upper=3> sd_mu_phi;
...
vector<lower=0, upper=1>[n_arms] mu_phi;
...
real<lower=0> omega_phi;
...
vector[Nind] eta_tilde_phi;
...
}
transformed parameters{
...
psi_phi=inv_logit(logit(mu_phi[arm_index])+eta_tilde_phi*omega_phi);
...
}
`
The hierarchical parameterization gives direct access to the 'typical value' of the parameter estimate, in the considered sample (ie. arm or study). In the nomemclature used in

PR #304, opened 2024-04-08
Plot Examples

gowerc:
No corresponding issue but this PR introduces example code to create the quantity plots that @mercifr1 was after. I think most are relatively straight forward.

The only exception is the population plot which requires some clunky manual derivation (with #296 having the potential to simplify this).


danielinteractive (approved):
super thanks @gowerc  - @mercifr1 does this do what you need?

Issue #170, opened 2023-07-06
Add additional links


danielinteractive:
To do:

- [ ] lm-random-slope:
	- [ ] "shared random effect joint model", which means just the patient specific deviation from the common slope as link contribution (scaled by link coefficient)
- [ ] GSF:
	- [ ] current longitudinal model value
	- [ ] growth rate parameter from the GSF
		- [ ] patient specific growth rate
		- [ ] individual deviation from population growth rate parameter
	- [ ] shrinkage rate parameter from the GSF
		- [ ] patient specific shrinkage rate
		- [ ] individual deviation from population shrinkage rate parameter

danielinteractive:
That is exactly correct

danielinteractive:
Hm that looks a bit clunky to me. I thought we have now methods for these links? I would rather like to see an API where each different link has another generic/method name.

danielinteractive:
Yes, I would recommend going this route. You can easily query using standard R methods stuff what works where.

Yes to your specific question.

danielinteractive:
Hm, I see... and we cannot / don't want to expose the true generics?... Just wondering if that would be long term easier to maintain as well

danielinteractive:
Hm I think it would be ok to pass `longmodel` as a user. It conceptually is well understandable.

danielinteractive:
I would still question whether this warrants the additional complexity...

danielinteractive:
You can decide, I guess we thought about it sufficiently

PR #308, opened 2024-04-14
Harmonise link methods

gowerc:
This update addresses the concerns raised in #170 about having 2 layers of link methods. Now there is only one e.g. `linkDSLD`, `linkIdentity` , `linkTTG`.

The functions have been designed such that the user can still call them without having to specify the longitudinal model object repeatedly e.g.

```
JointModel(
    survival = SurvivalWeibull(),
    longitudinal = LongitudinalRandomSlope(),
    link = Link(
        linkDSLD(prior = prior_normal()),
        linkTTG(prior = prior_gamma()),
        linkIdentity(prior = prior_normal())
    )
)
```

instead of

```
longmodel <- LongitudinalRandomSlope()
JointModel(
    survival = SurvivalWeibull(),
    longitudinal = longmodel,
    link = Link(
        linkDSLD(longmodel, prior = prior_normal()),
        linkTTG(longmodel, prior = prior_normal()),
        linkIdentity(longmodel, prior = prior_normal())
    )
)
```

This has been achieved by taking inspiration from JavaScript and having a `Promise` object. Essentially, by default the generics set their `model` argument to a `PromiseLongitudinal` object,  the corresponding method then returns a `PromiseLinkComponent` object which is essentially just a `LinkComponent` object but with a function slot instead of a `StanModule` slot.

The `JointModel` function then is responsible for resolving these promises by calling the function on the actual `LongitudinalModel` object that the user specifies.   I've posted a basic outline of how this works below:

So we first define the generic providing the default promise argument + a promise method that returns our promised link component which has a function to resolve the promise.
```
linkDSLD <- function(prior, model = PromiseLongitudinalModel(), ...) {
    UseMethod("linkDSLD", model)
}

linkDSLD.PromiseLongitudinalModel <- function(prior = prior_normal(0, 2), model, ...) {
    PromiseLinkComponent(fun = linkDSLD, prior = prior, key = "link_dsld")
}

linkDSLD.default <- function(prior, model, ...) {
    stop(sprintf("Method `linkDSLD` is not available for `%s`", class(model)[[1]]))
}
```

Then all we need to do is define our actual model method as normal (as the promise resolving function is just a call to the generic (essentially just delaying its execution). 

```
linkDSLD.LongitudinalGSF <- function(prior = prior_normal(0, 2), model, ...) {
    LinkComponent(
        key = "link_dsld",
        stan = StanModule("lm-gsf/link_dsld.stan"),
        prior = prior
    )
}
```

I think the code changes themselves aren't that complicated. Most of the files changed are really the documentation which I played around with quite a bit to try and make it simpler for end users. For example I added a "available links" section to each of the longitudinal models and I tried to re-write the "extending jmpost" documentation.

### DIFF
--- a/_pkgdown.yml
+++ b/_pkgdown.yml
@@ -173,6 +173,7 @@ reference:
       - Promise
       - PromiseLinkComponent
       - PromiseLongitudinalModel
+      - resolvePromise
       - resolvePromise.Link
       - resolvePromise.PromiseLinkComponent
   


danielinteractive (approved):
Thanks @gowerc , the documentation looks good. 

To be honest, overall, I am still not convinced between the tradeoff of small end user benefit (a little bit less code writing) and substantial code complexity in the package. Also in the future if maybe you want to have multiple longitudinal processes that influence the hazard of the survival model, it would be straightforward to do this in the standard way by calling the link on the longitudinal process it is attached to. 

But I leave it up to you, since you are the maintainer.

Issue #12, opened 2022-06-28
Understand how to predict new subjects in a mixed effect model


danielinteractive:
Background: What if we have a new patient in the test set that has a new random effect e.g. SLD parameters (e.g. ks)?

To do:
- [ ] think about methods side
  - [ ] if we have a much simpler linear mixed model e.g. fitted with lme4 package in R - how do the predictions work for new subjects? 
  - [ ] e.g. random intercept and random slope for a continuous covariate -> how does the prediction work?
- [ ] Consider the covariates that are modelled with random effects (corresponding to our SLD) as something we provide / have, then predict from that

gowerc:
@danielinteractive , Apologies if I'm mistaken but I thought the general practice of prediction with random effects was to just set the random effects to 0?  At least this makes intuiative sense to me as they are essentially nuisance parameters that are unique to the subject that you have no prior information on so the best you can do is assume the subject lies in the centre of the distribution (i.e. 0). 

I would have thought the only case where we wouldn't set to 0 is if we wanted to do simulation from the model (i.e. some form of say parametric bootstrapping).

danielinteractive:
Thanks @gowerc for looking at this - yeah generally I think the same. We just need to understand how it works here. Since here SLD is a covariate for the OS prediction. And I guess we need to somehow fit the SLD curve to the SLD observations of this individual to obtain meaningful OS predictions.

gowerc:
I feel like I'm potentially not understanding this properly :)  Wouldn't you just apply both models  i.e. take the model coeificients and use that to predict the patients SLD values based on their baseline covariates.  Then you can use the SLD values to predict the patients OS hazard and then convert that into S(t) =  exp (-H(t)) ?

I guess one challenge would be extracting the final OS hazard model from the stan code to calculate the hazard values. I think rstan has functions for exporting stan functions into R but I'm not sure if we have the same for cmdstan, perhaps we can use manipulate the stan code to create a stand alone stan program that simply takes OS inputs and returns the hazard values from day 0 - 1000 ?

danielinteractive:
Yeah exactly something like that

PR #313, opened 2024-05-10
Support for Survival Predictions

gowerc:
Partially addresses #12 

New api is:

```
surv_quantities <- SurvivalQuantities(
    mcmc_results,
    grid = GridPrediction(
        times = seq(1, 800, by = 50),
        newdata = dplyr::tibble(
            cov_cat = c("A", "B"),
            cov_cont = c(1.2, 2)
        ),
        params = list(
            intercept = 40,
            slope = 0.05
        )
    )
)
```

`newdata` allows  for multiple observations to be passed in but `params` only allows for 1 fixed set. Decided to go this route as the doing all combinations of `params` / `newdata` / `times` got a bit complicated & confusing. Users can query what `params` they need to define by running the `getPredictionNames()` helper function on the longitudinal model object.

### DIFF
--- a/R/GridPrediction.R
+++ b/R/GridPrediction.R
@@ -22,6 +22,24 @@ GridPrediction <- function(times = NULL, newdata, params = NULL) {
         newdata = newdata
     )
 }
+setValidity(
+    "GridPrediction",
+    function(object) {
+        for (param in names(object@params)) {
+            if (length(object@params[[param]]) != 1) {
+                return(sprintf("Parameter '%s' must be length 1", param))
+            }
+        }
+        if (length(object@params) != length(names(object@params))) {
+            return("All elements of `params` must be named")
+        }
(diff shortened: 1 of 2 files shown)


R/GridPrediction.R
```diff
@@ -0,0 +1,80 @@
...
+#' @export
+as.QuantityGenerator.GridPrediction <- function(object, data, ...) {
+
+    assert_class(data, "DataJoint")
+    data_list <- as.list(data)
+    validate_time_grid(object@times)
+
+    n_times <- length(object@times)
+    n_obs <- nrow(object@newdata)
+    newdata <- object@newdata
```
danielinteractive, line 36: unlikely to happen, but maybe safer to add an assertion here that this column name does not exist yet?

danielinteractive (dismissed):
thanks @gowerc , looks good

PR #330, opened 2024-05-21
Implements growth link function

gowerc:
Partially addresses #170 

Only thing that is making me a bit nervous about this is that the confidence intervals for this link parameter in my testing seem all most excessively wide.  For example with ~12 observations per subject and 800 subjects and a real value of 3 I'm getting:

```
# A tibble: 2 × 7
  variable       mean   q01   q99  rhat ess_bulk ess_tail
  <chr>         <dbl> <dbl> <dbl> <dbl>    <dbl>    <dbl>
1 link_growth   2.99  2.04   3.91  1.00    2035.    1249.
```

That might not seem so bad but then compare it to the other model parameters:

```
> dat
# A tibble: 5 × 7
  variable           mean    q01    q99  rhat ess_bulk ess_tail
  <chr>             <dbl>  <dbl>  <dbl> <dbl>    <dbl>    <dbl>
1 lm_sf_mu_bsld[1] 60.3   59.9   60.9    1.00    1895.     624.
2 lm_sf_mu_ks[1]    0.200  0.197  0.203  1.00    1879.    1311.
3 lm_sf_mu_ks[2]    0.251  0.248  0.254  1.00    1814.    1210.
4 lm_sf_mu_kg[1]    0.150  0.146  0.154  1.00    1691.     803.
5 lm_sf_mu_kg[2]    0.300  0.293  0.307  1.00    1770.    1302.
```

The confidence intervals are noticeably tighter. Admittedly this isn't really a unique problem for the growth link as it happens with all of them, but still makes me a little nervous... maybe its fine ? 

Either way the full link testing code takes ages to run so I've marked it as "not for cran" meaning it will only run in our twice a month CRON job.  I done a parser check though, e.g. it checks stans static parser to ensure the stan code is syntactically valid.

### DIFF
--- a/tests/testthat/test-LongitudinalSteinFojo.R
+++ b/tests/testthat/test-LongitudinalSteinFojo.R
@@ -348,7 +348,7 @@ test_that("Can recover known distributional parameters from a SF joint model wit
         c("lm_sf_mu_bsld", "lm_sf_mu_ks", "lm_sf_mu_kg"),
         TRUE
     )
-    true_values <- c(60, 0.2, 0.25, 0.15, 0.2)
+    true_values <- c(60, 0.2, 0.25, 0.15, 0.3)
     expect_true(all(dat$q01 <= true_values))
     expect_true(all(dat$q99 >= true_values))
     expect_true(all(dat$ess_bulk > 100))


danielinteractive (approved):
looks good, I think this just means that there is not a lot of information in the data about the link coefficient...

Issue #216, opened 2023-09-08
Harmonise Terminology of patients/subjects/observations/etc


gowerc:
Currently a mix of terminologies is used throughout the package name patients/subjects.  This should be made consistent.  That being said its not clear to me which of these to actually use, I would lean towards subject as it's broader but I don't know if there is a convention?

Likewise would we want to consider non-individual terms instead like observation / record / etc so that the package is in theory more generalisable as theres no reason it couldn't be used for non-trial data analysis.

danielinteractive:
Yeah "subject" is good as a convention here, similar "observation" sounds good to me

mercifr1:
For the "patient"-level, I would also be in favor of "subject".
For the "record"-level, I would also prefer the term "observation".

PR #331, opened 2024-05-22
Convert "patient" -> "subject"

### LINKED ISSUE #325: Replace references to "pt" in the simulation functions to "subject"
Lots of references still to patient in the simulation functions, these should all be replaced with sub or subject. Though this might be quite painful given how prolific they are...

gowerc:
Closes #325  and #216 

Basically a mass find and replace of pt / patient -> subject.  Had to go through it all manually as there was a lot of "excePT" and "intercePT" which made a literal find and replace unviable :(

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -83,11 +83,11 @@ inst/stanmodels/stan_model
 *.log
 
 
+# Local R environment variables
+.Renviron
+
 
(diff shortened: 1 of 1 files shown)


danielinteractive (approved):
super, thanks @gowerc

PR #343, opened 2024-06-06
Remove model cache

gowerc:
Currently every other model run on GitHub actions is [failing](https://github.com/Genentech/jmpost/actions/runs/9398586560/job/25884227228) due to what I think is an issue with the GitHub cache. Theres no obvious reason as to why this is happening so will need to investigate at a later date. For now I'm just going to disable the GitHub cache so that the models are recompiled from scratch each time.

### DIFF
--- a/.github/workflows/check.yaml
+++ b/.github/workflows/check.yaml
@@ -31,6 +31,10 @@ jobs:
         CMDSTAN=/root/.cmdstan
         CMDSTAN_PATH=/root/.cmdstan
         CMDSTANR_NO_VER_CHECK=true
+      ### Temporarily removed due to cache bug (#345)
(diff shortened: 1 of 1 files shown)


.github/workflows/check.yaml
```diff
@@ -31,9 +31,6 @@ jobs:
         CMDSTAN=/root/.cmdstan
         CMDSTAN_PATH=/root/.cmdstan
         CMDSTANR_NO_VER_CHECK=true
-        JMPOST_CACHE_DIR=${{ github.workspace }}/.cache
```
danielinteractive, line 34: maybe comment rather than delete to simplify reactivation later?

danielinteractive (approved):

PR #344, opened 2024-06-06
updated priors for regression coeficients

### LINKED ISSUE #341: Set Survival regression priors of the beta coeficients to N(0,2)
Current prior of N(0,5) is considered to broad and thus agreed to reduce to N(0,2) which I think is considered a more standard "uninformative prior" for log-regression coeficients

gowerc:
Closes #341


danielinteractive (approved):
LGTM, any docs which need to be updated?

PR #350, opened 2024-06-07
Claret Bruno Model

### LINKED ISSUE #338: Implement Claret-Bruno Longitudinal Model
Original details from @mercifr1  (re-worded slightly to make sense outside of the context of the initial email)

---

The 'Claret-Bruno' model was called the "simplified TGI" model by Laurent Claret, in contrast to the alternative model which would include a drug exposure term. We are interested in the former ("simplified"), without drug exposure.

The term "simplified TGI" appears in the 2015 CCP paper (2), and refers to the 2013 JCO paper (1) (ref. 23 and not 22!) which is the first publication presenting this model. In the Supplementary material of the 2013 JCO paper (1), you will find a model implementation including drug exposure, expressed either as ODE or in analytical form. Setting Exposure to 1, you obtain the "simplified TGI" model.

With more TGI models being published, the "simplified TGI" model was renamed and is now called the 'Claret-Bruno' model in the Pharmacometrics community.

In these two papers, the model formulation is not accurate, unfortunately. For instance, what happens when t<0 is not specified. A recent paper (2023 Bmtx (3)) using the Claret-Bruno model in equation [1] (section 2.2), has more accurate formulation.

References:

1) 10.1200/JCO.2012.45.0973
2) 10.1007/s00280-015-2820-x
3) 10.1111/biom.13912

gowerc:
Closes #338 

Most details are in the linked issue. Otherwise not too much special here just based most off the code off of the existing SF model with no real fundamental changes. 

Only concern I have is that the ESS seems to be quite low at ~10%.  Though this model has a double exponential and some terms that are hard to pick a part so I don't think its a validity concern, its not like the old non-central b parameter that only had single figure ess even with thousands of samples....  but yer not 100% sure that this is ok to ignore but my gut feeling says it is. 

EDIT - Talking to Francois he was saying that the SB model tends to have slightly worse performance metrics than the GSF which means we are roughly in line with this expectation so I think this is safe to ignore.

### DIFF
--- a/inst/stan/base/functions.stan
+++ b/inst/stan/base/functions.stan
@@ -82,32 +82,50 @@ functions {
         return ret;
     }
 
-    // Replaces the values of `y` with `replacement` if the corresponding value
-    // of `x` is less than 0.
-    matrix if_gte0_else(matrix x, matrix y, real replacement) {
-        matrix[rows(x), cols(x)] result;
-        for (i in 1:rows(x)) {
-            for (j in 1:cols(x)) {
-                result[i, j] = x[i, j] >= 0 ?  y[i, j] : replacement;
+    // Element wise "if greater than 0 return y, else return y"
+    //
+    // Input:
+    //     item: a matrix or vector
+    //     x: an object of the same type and dimensionality of `item`
+    //     y: an object of the same type and dimensionality of `item` or a `real`
+    //
+    // Description:
+    //     The following overloaded functions take an object "item"
+    //     and loop over each element returning the corresponding element
+    //     of `x` if the element is >= 0 else returning the corresponding
+    //     element of `y` (or just `y` if why is a single real)
+    //
+    // Returns:
+    //     an object of the same type and dimensionality of `item`
+    //     but with the values of either `x` or `y`. 
+    vector if_gte0_else(vector item, vector x, real y) {
+        vector[rows(item)] result;
+        for (i in 1:rows(item)) {
+            result[i] = item[i] >= 0 ? x[i] : y;
+        }
+        return result;
+    }
+    vector if_gte0_else(vector item, vector x, vector y) {
+        vector[rows(item)] result;
+        for (i in 1:rows(item)) {
+            result[i] = item[i] >= 0 ? x[i] : y[i];
+        }
+        return result;
+    }
+    matrix if_gte0_else(matrix item, matrix x, real y) {
+        matrix[rows(item), cols(item)] result;
+        for (i in 1:rows(item)) {
+            for (j in 1:cols(item)) {
+                result[i, j] = item[i, j] >= 0 ?  x[i, j] : y;
             }
         }
         return result;
     }
-    vector if_gte0_else(vector x, vector y, real replacement) {
-        vector[rows(x)] result;
-        for (i in 1:rows(x)) {
-            result[i] = x[i] >= 0 ? y[i] : replacement;
-        }
-        return result;
-    }
-
-    matrix if_gte0_else(matrix x, matrix y, matrix replacement) {
-        int nrow = rows(x);
-        int ncol = cols(x);
-        matrix[nrow, ncol] result;
-        for (i in 1:nrow) {
-            for (j in 1:ncol) {
-                result[i, j] = x[i, j] < 0 ? replacement[i, j] : y[i, j];
+    matrix if_gte0_else(matrix item, matrix x, matrix y) {
+        matrix[rows(item), cols(item)] result;
+        for (i in 1:rows(item)) {
+            for (j in 1:cols(item)) {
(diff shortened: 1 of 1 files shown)


vignettes/statistical-specification.Rmd
```diff
@@ -354,6 +354,106 @@ of the distribution is used. This is because this has the same interpretation as
...
+\right)
+exp \left(
+    g_i t_{ij}
+    \frac{
+        p_i \left(1 - e^{-c_i t_{ij}}\right)
+    }
+    {c_i}
+\right)
+ & \text{if } t\geq 0 \\
+b_i g_i e^{g_i t_{ij}} & \text{if } t\lt 0
```
danielinteractive, line 411:
```suggestion
 & \text{if } t_{ij}\geq 0 \\
b_i g_i e^{g_i t_{ij}} & \text{if } t_{ij}\lt 0
```
and similarly in other places

vignettes/statistical-specification.Rmd
```diff
@@ -354,6 +354,106 @@ of the distribution is used. This is because this has the same interpretation as
...
+* $g_i$ is the subject tumour growth rate.
+* $c_i$ is the subject treatment resistance rate.
+* $p_i$ is the subject treatment growth inhibition response.
+* $k(i)$ is the treatment arm index for subject $i$
+* $l(i)$ is the study index for subject $i$
+* $\mu_{\theta k(i)}$ is the population mean for parameter $\theta$ in group $k(i)$
+* $\omega_{\theta}$ is the population variance for parameter $\theta$.
+
+
+### Derivative of the SLD Trajectory Link
```
danielinteractive, line 395: I double checked this formula and it is correct, one idea is that for simplicity you could reuse the `SLD(t_{ij})` expression in the formulas.

inst/stan/lm-claret-bruno/link_dsld.stan
```diff
@@ -0,0 +1,42 @@
...
+        );
+
+        matrix[nrows, ncols] result_gt0 = fmin(
+            8000.0,
+            ind_b_matrix .*
+            (
+                ind_g_matrix - ind_p_matrix .* exp(-ind_c_matrix .* time)
+            ) .*
+            exp(
+                (ind_g_matrix .* time) .*
```
danielinteractive, line 35:
```suggestion
                (ind_g_matrix .* time) .-
```
I think this was a typo here, hopefully that will fix the problem

inst/stan/lm-claret-bruno/link_dsld.stan
```diff
@@ -0,0 +1,42 @@
...
+                (ind_p_matrix ./ ind_c_matrix) .*
+                (1 -  exp(- ind_c_matrix .* time))
+            )
+        );
+        return if_lt0_else(time, result_gt0, result_lt0);
```
danielinteractive, line 40:
```suggestion
        return if_lt0_else(time, result_lt0, result_gt0);
```
? (at least from reading the function name this would be logical)

inst/stan/lm-claret-bruno/link_dsld.stan
```diff
@@ -0,0 +1,42 @@
...
+        vector[nrows] ind_c = link_function_inputs[,3];
+        vector[nrows] ind_p = link_function_inputs[,4];
+        
+        // Here we assume that psi's are replicated along the rows of the time matrix.
+        matrix[nrows, ncols] ind_b_matrix = rep_matrix(ind_b, ncols);
+        matrix[nrows, ncols] ind_g_matrix = rep_matrix(ind_g, ncols);
+        matrix[nrows, ncols] ind_c_matrix = rep_matrix(ind_c, ncols);
+        matrix[nrows, ncols] ind_p_matrix = rep_matrix(ind_p, ncols);
+
+        matrix[nrows, ncols] result_lt0 = fmin(
```
danielinteractive, line 23: unrelated idea - it would be better to centralize these `fmin(8000.0, .)` calls in something like `safecap()` such that the hardcoded 8000 value could be replaced in one go, and in order to enhance readability

danielinteractive (commented):
Thanks @gowerc , I think I found something

danielinteractive:
Great news!! Thanks both!!

danielinteractive:
Hi @gowerc , yeah sure, anything specific to check still?

inst/stan/base/functions.stan
```diff
@@ -100,5 +100,17 @@ functions {
         }
         return result;
     }
+
```
danielinteractive, line 103: maybe add comments on the function

danielinteractive (approved):
thanks, otherwise looks good

danielinteractive (approved):
thanks Craig!

PR #354, opened 2024-06-14
Harmonise Simulation and formula parameter values

### LINKED ISSUE #349: Double check initial values for SF and GSF models
Whilst developing the Claret-Bruno model I noticed that we had a few cases where  exp(values) were being passed to places where they should have been logged. Just double check to see if these models are having the same issue. Particularly with regards to the population means

gowerc:
Closes #349

Bit of a deviation from the original issue. The initial values are indeed correct but there was a bit of inconsistency that the model and formulas were specifying the raw value but the simulation functions were specifying the exp(value). I've updated the simulation functions to specify the raw value so that everything is consistent.

Sorry to re-explain that I originally thought the initial values were wrong because they were so wildly different to the values I was specifying in the simulation function. But they aren't wrong they are just on different scales which is confusing so I have harmonised the scales.


vignettes/extending-jmpost.Rmd
```diff
@@ -540,9 +540,9 @@ SimLongitudinalGSF <- function(
     sigma = 0.01,
     mu_s = c(0.6, 0.4),
     mu_g = c(0.25, 0.35),
-    mu_b = 60,
-    a_phi = c(4, 6),
-    b_phi = c(4, 6),
+    mu_b = log(60),
+    a_phi = log(c(4, 6)),
+    b_phi = log(c(4, 6)),
```
danielinteractive, line 545: correct?

PR #361, opened 2024-06-19
Add toggle for Graph Snapshot Testing

gowerc:
No corresponding issue.

Unit tests were failing on internal CICD build due to snapshot discrepancies, as far as I can tell this is minor white space stuff on graphs due to differences in package versions (I've posted the diff below). To address this I've added a toggle for graph snapshots which is enabled for the GitHub CICD but disabled for the in house builds. 

I opted for a wrapper around the `expect*` function rather than skipping the test as this way it will still through an error on the internal build if the plot generation code fails its only the final comparison that is suppressed. 

```
> diff  \
    survival-plot-no-wrap-no-ci-km-ggplot2.new.svg  \
    survival-plot-no-wrap-no-ci-km-ggplot2.svg
    
21c21
< <rect x='0.00' y='0.00' width='720.00' height='576.00' style='stroke-width: 1.07; stroke: #FFFFFF; fill: #FFFFFF;' />
---
> <rect x='0.000000000000064' y='0.00' width='720.00' height='576.00' style='stroke-width: 1.07; stroke: #FFFFFF; fill: #FFFFFF;' />
````

### DIFF
--- a/R/JointModelSamples.R
+++ b/R/JointModelSamples.R
@@ -104,7 +104,7 @@ as_print_string.JointModelSamples <- function(object, indent = 1, ...) {
     sprintf(
         paste(template_padded, collapse = "\n"),
         as.CmdStanMCMC(object)$metadata()$iter_sampling,
-        as.CmdStanMCMC(object)$metadata()$num_chains
+        as.CmdStanMCMC(object)$num_chains()
     )
 }
 
--- a/tests/testthat/_snaps/JointModelSamples.md
+++ b/tests/testthat/_snaps/JointModelSamples.md
@@ -7,7 +7,7 @@
          JointModelSamples Object with:
         
             # of samples per chain = 100
-            # of chains            = 1
+            # of chains            = 2
         
             Variables:
                 Ypred[2800]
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

File: tests/testthat/test-survival_plot.R [hunk 1]
```diff
@@ -63,10 +64,13 @@ test_that("survival_plot works as expected", {
         x_label = expression(thd[3])
     )
 
-    vdiffr::expect_doppelganger(
-        "survival_plot with wrap and ci",
-        p1
-    )
+    if (is_graph_snapshot_enabled()) {
```
gravesti, line 67: Would it be nicer to use one of the conditional skip functions? eg
```suggestion
    skip_if_not(is_graph_snapshot_enabled()) {
```

gravesti (commented):
Just one suggestion. Maybe you already considered it? @gowerc

tests/testthat/test-survival_plot.R
gravesti: You're right, I thought it could be done for a block.

gravesti (approved):

PR #362, opened 2024-06-19
Fix number of chains print method

### LINKED ISSUE #356: Prints incorrect number of chains for JointModelSamples
Printing a JointModelSamples object appears to have the wrong number of chains

```r
mp <-  sampleStanModel(
            jm,
            data = jdat,
            iter_sampling = 1000,
            iter_warmup = 1000,
            chains = 3,
            parallel_chains = 4
        )
mp
```
prints
```
   JointModelSamples Object with:
  
      # of samples per chain = 1000
      # of chains            = 1
  
      Variables:
          beta_os_cov[3]
          link_coefficients
          ...
```
but I expect number of chains should be 3.

This comes from
https://github.com/Genentech/jmpost/blob/c29f73f63ffe693220741358e02f80c86e5f7180/R/JointModelSamples.R#L107
and I think it should be
```r
 as.CmdStanMCMC(object)$num_chains()
```
eg
```r
as.CmdStanMCMC(mp)$num_chains()
[1] 3
```

gowerc:
Closes #356 

Minor side effect was that the main test object I have (which is shared across many tests) only used 1 chain so I had to increase that to 2 for the snapshots to show this change is working; but by doing so it impacted some other snapshot tests  which I had to update accordingly.

### DIFF
--- a/R/JointModelSamples.R
+++ b/R/JointModelSamples.R
@@ -104,7 +104,7 @@ as_print_string.JointModelSamples <- function(object, indent = 1, ...) {
     sprintf(
         paste(template_padded, collapse = "\n"),
         as.CmdStanMCMC(object)$metadata()$iter_sampling,
-        as.CmdStanMCMC(object)$metadata()$num_chains
+        as.CmdStanMCMC(object)$num_chains()
     )
 }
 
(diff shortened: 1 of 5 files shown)


gravesti (approved):
Nice! @gowerc

PR #364, opened 2024-06-19
Defer inclusion of generated quantity code

### LINKED ISSUE #322: Defer generated quantity code
Currently all generated quantity code must be provided in advance at the time of initial model fitting. The package then nulls out this code so that it isn't run during the initial sampling.

Could potentially be a nicer user interface if we could defer the provision of this code to the when the generated quantities are actually called. Potentially this could be done via a dedicated method like `enableGeneratedQuantities()` the same we we currently have the `enableLink()` method for injecting any link specific code.

gowerc:
Closes #322 

General gist is that all generated quantity code had to be defined in advance even though its not used/compiled until the post processing.  This was slightly awkward that it meant the user can't alter the GQ code without re-compiling  / re-sampling the whole model and it was in complete contrast to how we handle the link code.  This PR attempts to rectify this. All round this isn't that much of a feature change but hopefully just makes the package a tiny bit easier to use with some more separation of concerns.

### DIFF
--- a/vignettes/extending-jmpost.Rmd
+++ b/vignettes/extending-jmpost.Rmd
@@ -290,7 +290,7 @@ enableGQ.LongitudinalRandomSlope <- function() {
 }
 ```
 
-Note that whilst it is possible to provide a `enableGQ()` method for the survival model it is not
+Note that whilst it is possible to provide an `enableGQ()` method for the survival model it is not
 required. This is because the underlying framework for creating survival quantities
 is distribution agnostic and does not require any model specific code.
 


gravesti (commented):
@gowerc Looks ok to me. Should there any new tests or changes to existing tests?

vignettes/extending-jmpost.Rmd
```diff
@@ -241,6 +247,46 @@ as these are handled by the `parameters` slot of the model object as mentioned a
...
 
+
+## Custom Generated Quantities
+
+In order to avoid unnecessary processing, code that is only used for generated quantities is
+only included into the Stan model when explicitly trying to extract the quantities via the
+`LongitidunalQuantities` and `SurvivalQuantities` constructors.
+
+To facilitate this, when using `LongitidunalQuantities` or `SurvivalQuantities` 
+a dedicated model method `enableGQ()` is called on the longitudinal model. This model specific
```
gravesti, line 258: Generally the docs are clear. @gowerc 
This bit stuck out: If `SurvivalQuantities` is used does it really run `enableGQ()` on the longitudinal model (only)?

vignettes/extending-jmpost.Rmd
```diff
@@ -241,6 +247,54 @@ as these are handled by the `parameters` slot of the model object as mentioned a
...
+
+            matrix[gq_n_quant, 2] long_gq_pop_parameters;
+            long_gq_pop_parameters[, 1] = to_vector(lm_rs_intercept[gq_long_pop_study_index]);
+            long_gq_pop_parameters[, 2] = to_vector(lm_rs_slope_mu[gq_long_pop_arm_index]);
+        }
+    ")
+}
+```
+
+Note that whilst it is possible to provide a `enableGQ()` method for the survival model it is not
```
gravesti, line 293:
```suggestion
Note that whilst it is possible to provide an `enableGQ()` method for the survival model it is not
```

gravesti (dismissed):
Docs read better now @gowerc
I just spotted one typo to fix

gravesti (approved):

PR #366, opened 2024-06-21
Stop deletion of snapshot files

### LINKED ISSUE #365: Deleting Unused Snapshots
Appears to be an unintended side effect of #361 in that when the snapshots are note used they are deleted... this could easily cause issues if the deletions are pushed back to the repo.

```
Deleting unused snapshots:
• survival_plot/survival-plot-no-wrap-no-ci-km-ggplot2.svg
• survival_plot/survival-plot-with-no-wrap-and-no-ci.svg
• survival_plot/survival-plot-with-wrap-and-ci.svg
```

gowerc:
Closes #365 

I dare not admit to how much time I spent on this because I didn't realise the snapshot file generation was automatically converting `_` -> `-` without me realising ...  😠

### DIFF
--- a/tests/testthat/test-survival_plot.R
+++ b/tests/testthat/test-survival_plot.R
@@ -61,14 +61,16 @@ test_that("survival_plot works as expected", {
         get_data(3, "C")
     )
 
-    p1 <- survival_plot(
-        res,
-        add_ci = TRUE,
-        add_wrap = TRUE,
-        kmdf = NULL,
-        y_label = expression(frac(1, 2) + S(t^2)),
-        x_label = expression(thd[3])
-    )
+    testthat::expect_no_error({
+        p1 <- survival_plot(
+            res,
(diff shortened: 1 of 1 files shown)


tests/testthat/test-survival_plot.R
```diff
@@ -100,10 +109,15 @@ test_that("survival_plot works as expected", {
...
         vdiffr::expect_doppelganger(
-            "survival_plot-no wrap + no ci + km + ggplot2 ",
+            "nowrap-noci-km-ggplot2",
             p3
         )
     }
+
+    # Null test so that its not considered as an empty test if the "JMPOST_GRAPH_SNAPSHOT"
+    # variable is not set
+    expect_equal(1, 1)
```
gravesti, line 122: instead you wrap the `p3 <- survival_plot(`  etc in `expect_no_error()`

gravesti (approved):

PR #370, opened 2024-06-25
Fix SF Quantities

### LINKED ISSUE #368: SF code refers to GSF variables
See - https://github.com/Genentech/jmpost/blob/main/inst/stan/lm-stein-fojo/quantities.stan  Should be SF

gowerc:
Closes #368 

Ok so the actual bug fix was trivial was just a case of fixing a simple typo  e.g. `gsf` -> `sf`. What was a bit more involved though was updating the unit tests to actually catch this error.   It would have been easy to just compile all of the models but this would have added a ton of run time especially if we want to test each permutation of model / quantity generation.

Instead I decided to abstract the code that generates the `StanModule` for the quantities into its own function so that we can just grab that object before its actually compiled. I then put together a mock `JointModelSamples` object to which I then use to generate the corresponding `StanModule` and then run the stan parser to make sure theres no obviously variable errors. 

I then added the unit test to our other longitudinal models to make sure no other bugs like this exist. 

A bit of a red herring in these changes was that my original implementation tried to expand on the `write_stan()` function. I adapted the function to take a connection stream to which  I was going to pipe the output to a  variable. I abandoned this implementation for testing this PR but I thought having the function take a stream was a nice change so left it in as I had already written the code (hope thats ok).

### DIFF
--- a/R/JointModel.R
+++ b/R/JointModel.R
@@ -142,7 +142,7 @@ as.character.JointModel <- function(x, ...) {
 
 #' @rdname write_stan
 #' @export
-write_stan.JointModel <- function(object, destination) {
+write_stan.JointModel <- function(object, destination, ...) {
     if (is_connection(destination)) {
         return(writeLines(as.character(object), con = destination))
     }
--- a/R/SimLongitudinalClaretBruno.R
+++ b/R/SimLongitudinalClaretBruno.R
@@ -175,7 +175,7 @@ sampleSubjects.SimLongitudinalClaretBruno <- function(object, subjects_df) {
 
 #' Claret-Bruno Functionals
 #'
-#' @param time (`numeric`)\cr time grid.
+#' @param t (`numeric`)\cr time grid.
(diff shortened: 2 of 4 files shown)


R/JointModel.R
```diff
@@ -142,8 +142,11 @@ as.character.JointModel <- function(x, ...) {
 
 #' @rdname write_stan
 #' @export
-write_stan.JointModel <- function(object, file_path) {
-    fi <- file(file_path, open = "w")
+write_stan.JointModel <- function(object, destination) {
+    if (is_connection(destination)) {
+        return(writeLines(as.character(object), fi))
```
gravesti, line 147: Something like this, right?
```suggestion
        return(writeLines(as.character(object), con = destination))
```

gravesti (commented):
Looks good to me, @gowerc 
just one bug I think

gravesti (dismissed):

gravesti (approved):

PR #372, opened 2024-06-27
Support initial values for constrained parameters

### LINKED ISSUE #268: Document that users need to set their own initial values for truncated distributions
A common pattern in Stan is to use unconstrained distributions but then simply specify a constraint band within the stan code.

For example

```
real<lower=0> sigma;
sigma ~ cauchy(0, 5);
```

Our current method for initial value generation is to sample one from the prior distribution, however this doesn't currently respect the constraints placed on the value within the stan code which can lead to the model failing on the first pass e.g.

```
Running MCMC with 1 chain...

Chain 1 Rejecting initial value:
Chain 1   Error evaluating the log probability at the initial value.
Chain 1 Exception: lb_free: Lower bounded variable is -0.47888, but must be greater than or equal to 0.000000 (in '/var/folders/hs/gyg0q5g94pz917klnkg7tnt80000gq/T/RtmpXZWCT7/model-153173ae186df.stan', line 181, column 4 to column 45)
Chain 1 Exception: lb_free: Lower bounded variable is -0.47888, but must be greater than or equal to 0.000000 (in '/var/folders/hs/gyg0q5g94pz917klnkg7tnt80000gq/T/RtmpXZWCT7/model-153173ae186df.stan', line 181, column 4 to column 45)
```

Couple of questions arise from this:

- Is this a bug that needs to be fixed? As in if users are using unconstrained distributions for constrained parameters then perhaps we can push onus on to them to define suitable starting values?
- That being said this is a recommended way of working with Stan so maybe we need to take accountability for this
- If we need to take accountability should this be done at the user level or behind the scenes (e.g. we update their priors with the known contraints or are the users responsible for specifying the constraint in the prior?)
- I'm not even sure how to implement this... I'm guessing an easy way would just be for >0 constraint and then simply `abs()` the sampled value? but what about arbitrary constraints ?

gowerc:
Closes #268

Generally tried to shield the user from this as much as possible and pushed all code to the behind the scenes. Each model (e.g. `LongitudinalGSF`) is responsible for setting the constraints on the priors which is done via `set_limits()`. 

`initialValues()` then attempts to draw 100 values and discards any that do not meet the constraint. If there is more than 1 value that meets the constraint then 1 of these is sampled at random. 

For random effects parameters that were having their distribution set based on the initial value of the parent I've tried to reduce the variance of the values being produced by implementing a `median` function. In short this just generates 250 values following the above described method and then just takes the median across these.  Originally we were just taking the centre point of the distribution but as users can now specify distributions whose centre point is potentially not a valid value this alternative was required. 

Misc changes
- Changed the `prior@init` slot to `prior@centre` which better represents what this parameter is doing (its meant to denote the centre of the distribution or more accurately the point at which we shrink our initial values towards).
- Reduced the default prior of the longitudinal mu slope. As the new initial value selection has more noise / variance to it  it started to generate values for the random effects that caused initial sampler errors so tightened the default prior to help reduce this
- For reasons I don't understand the shared environment `test_data_01` which is generated by a helper function started throwing errors when the unit tests were run in parallel. I've had to change the code slightly such that the environment is now individually created in each file that needs it. This does unfortunately add some run time but I couldn't find another solution :(

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -25,7 +25,7 @@ License: Apache License (>= 2)
 Encoding: UTF-8
 Language: en-GB
 Roxygen: list(markdown = TRUE)
-RoxygenNote: 7.3.1
+RoxygenNote: 7.3.2
 Depends:
     R (>= 4.1.0)
 Imports:


R/Prior.R
```diff
@@ -52,28 +54,31 @@ NULL
 #' @param repr_model (`string`)\cr the Stan code representation for the model block.
 #' @param repr_data (`string`)\cr the Stan code representation for the data block.
 #' @param display (`string`)\cr the string to display when object is printed.
-#' @param init (`numeric`)\cr the initial value.
+#' @param centre (`numeric`)\cr the central point of distribution to shrink sampled values towards
 #' @param validation (`list`)\cr the prior distribution parameter validation functions. Must have
 #' the same names as the `paramaters` slot.
 #' @param sample (`function`)\cr a function to sample from the prior distribution.
+#' @param limits (`numeric`)\cr TODO
```
gravesti, line 61: the lower and upper limits for a truncated distribution.

R/Prior.R
```diff
@@ -7,7 +7,7 @@ NULL
 #' The documentation lists all the conventional arguments for [`Prior`]
 #' constructors.
 #'
-#' @param init (`number`)\cr initial value.
+#' @param centre (`number`)\cr the central point of distribution to shrink sampled values towards
```
gravesti, line 10: what exactly is the central point? the median?

gravesti (commented):
@gowerc A few things to clarify :)

gravesti (dismissed):

gravesti (approved):
very nice @gowerc !

PR #375, opened 2024-06-28
Logit-Normal parameterisation for GSF Phi

### LINKED ISSUE #374: Parameterization of GSF phi parameter
A constant request from @mercifr1 is to use the logit-normal (see Kerioui et al. 2020 SiM) parameterisation for this parameter.  Need to investigate the consequences of doing this with respect to the centered and non-centered parameterisations

gowerc:
Closes #374 

A bit messy just because the GSF distribution is used all over the place so had to edit a lot of testing locations. 

The implementation is slightly different compared to the other parameters.  This is because Stan doesn't have a native logit-normal distribution. Some of their articles recommend implementing logit-normal via:

```
logit(phi) ~ normal(mu, sigma)
```
Problem is if you do it this way you need to apply a Jacobian adjustment which I didn't want to mess around with. So to avoid that instead of having `phi` as the parameter I have `phi_logit` as the parameter (thus placing the prior on `phi_logit`) and then transform it to the required  value via `phi = inv_logit(phi_logit)`

### DIFF
--- a/R/LongitudinalGSF.R
+++ b/R/LongitudinalGSF.R
@@ -36,8 +36,8 @@ NULL
 #' @param omega_bsld (`Prior`)\cr for the baseline value standard deviation `omega_bsld`.
 #' @param omega_ks (`Prior`)\cr for the shrinkage rate standard deviation `omega_ks`.
 #' @param omega_kg (`Prior`)\cr for the growth rate standard deviation `omega_kg`.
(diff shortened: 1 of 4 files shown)


gravesti (commented):
One suggestion to improve the docs, otherwise looks good @gowerc

gravesti (approved):

PR #397, opened 2024-07-31
use cmdstanr generic

### LINKED ISSUE #387: Potential export issue with `as.CmdStanMCMC`
Not 100% sure whats going on but looks like there might be a Namespace issue with as.CmdStanMCMC.

```
devtools::load_all(export_all = FALSE)

Error in UseMethod("as.CmdStanMCMC") : 
  no applicable method for 'as.CmdStanMCMC' applied to an object of class "JointModelSamples"
 ```

### PR DESCRIPTION
**Pull request by gowerc**
Closes #387

### DIFF
--- a/R/zzz.R
+++ b/R/zzz.R
@@ -18,7 +18,7 @@
 
         if (length(possible_paths)) {
             for (try_path in possible_paths) {
-                new_path <- tryCatch( 
+                new_path <- tryCatch(
                     suppressMessages(cmdstanr::set_cmdstan_path(try_path)),
                     warning = function(w) NULL,
                     error = function(e) NULL

### CODE REVIEW

**Comment by gravesti** [comment]
Wonder if something like this could work ?
https://forum.posit.co/t/write-s3-methods-for-generics-from-other-packages-without-importing-that-package/8884

File: R/JointModelSamples.R [hunk 1]
```diff
@@ -98,7 +98,7 @@ as.StanModule.JointModelSamples <- function(object, generator, type, ...) {
 #' @export
 as_print_string.JointModelSamples <- function(object, indent = 1, ...) {
     sizes <- vapply(
-        as.CmdStanMCMC(object)$metadata()[["stan_variable_sizes"]],
+        cmdstanr::as.CmdStanMCMC(object)$metadata()[["stan_variable_sizes"]],
```
gravesti, line 101: Is this now unconditional use of a suggested package?

gravesti (approved):

PR #408, opened 2024-08-14
Individual Loo for submodels 

### LINKED ISSUE #402: Update loo implementation to be for individual sub models only
Following on from #391  it transpires that loo implementation doesn't make sense for the full joint model (as OS is done on the subject level whereas tumour data is on an observation level so what does it mean to remove an observation?).  Thus we agreed to support loo only for individual sub models and not the full joint model.

gowerc:
Closes #402


gravesti (dismissed):

tests/testthat/test-brms_compare.R
```diff
@@ -0,0 +1,251 @@
...
+        chains = 2,
+        refresh = 200,
+        parallel_chains = 2
+    )
+
+
+    stanmod <- as.CmdStanMCMC(mp)
+
+
+    mp_brms <- brms::brm(
```
gravesti, line 104: I guess strictly speaking these tests should have a `skip_if_not(requireNamespace('brms')`
@gowerc

gravesti:
@gowerc I think you might not have pushed your latest changes yet :)

gravesti (approved):

tests/testthat/test-brms_compare.R
gravesti: As discussed, leave as is

gravesti (approved):

PR #425, opened 2025-01-30
Add brms examples for non-linear model

### LINKED ISSUE #417: Add brms example code that uses multiple studies

gowerc:
Closes #417

Also moved a bunch of older stuff that I don't think is relevant anymore to a "legacy" folder to clean up the design area a bit

### DIFF
--- a/design/examples/brms_nonlin_mixeff.R
+++ b/design/examples/brms_nonlin_mixeff.R
@@ -171,7 +171,14 @@ ggplot(dat_plot, aes(x = est_value, y = value)) +
 
 set.seed(283)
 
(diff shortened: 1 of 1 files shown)


gravesti:
@gowerc For the brms examples I noticed the estimates are not near the diagonals like the jmpost ones are. Is that expected?
### jmpost model
image

### brms model
image

gravesti:
Thanks @gowerc Maybe worth a note in the script for posterity to say that the models are sensitive to initial values are require a good warmup phase?

gravesti (approved):

PR #430, opened 2025-01-31
Support truncated prior distributions

### LINKED ISSUE #429: Truncation missing for standard deviation parameters?
Hi @gowerc , @mercifr1 , @gravesti 

When comparing `jmpost` with `brms`, I noticed a difference in the generated Stan code that might be relevant.
When using exactly the same priors in both (where this matters), I still get slightly different results, and I wonder if this is the reason, or at least a contributing factor.

- In `jmpost`, when declaring a normal prior for a standard deviation, the underlying Stan code still says e.g.
  ```
  lm_sf_omega_bsld ~ normal(prior_mu_lm_sf_omega_bsld, prior_sigma_lm_sf_omega_bsld);
  ```
- In `brms`, when doing the same, the underlying Stan code contains code that takes into account the truncation boundary by using the log cdf, e.g.:
  ```
  lprior += normal_lpdf(sd_3 | 0, 3)
    - 1 * normal_lccdf(0 | 0, 3);
  ```

So I checked the Stan manual, and found this:
https://mc-stan.org/docs/reference-manual/statements.html#truncation-with-lower-bounds-in-stan
Based on this, I would assume that the Stan code generated by `jmpost` should actually read:

```
lm_sf_omega_bsld ~ normal(prior_mu_lm_sf_omega_bsld, prior_sigma_lm_sf_omega_bsld) T[0, ];
```

Did you come across this maybe already?

gowerc:
Closes #429 

FYI @danielinteractive

### DIFF
--- a/R/Prior.R
+++ b/R/Prior.R
@@ -150,7 +150,7 @@ as.character.Prior <- function(x, ...) {
 
 
 #' Creates Stan Syntax for Truncated distributions
-#' @description 
+#' @description
 #' This function creates the Stan syntax for truncated distributions
 #' @param limits (`numeric`)\cr the lower and upper limits for a truncated distribution
 #' @keywords internal


tests/testthat/_snaps/LongitudinalSteinFojo.md
```diff
@@ -29,10 +29,10 @@
           lm_sf_mu_bsld ~ normal(mu = 4.09434, sigma = 1)
           lm_sf_mu_ks ~ normal(mu = -0.69315, sigma = 1)
           lm_sf_mu_kg ~ gamma(alpha = 2, beta = 1)
-          lm_sf_omega_bsld ~ lognormal(mu = -1.60944, sigma = 1)
-          lm_sf_omega_ks ~ lognormal(mu = -1.60944, sigma = 1)
-          lm_sf_omega_kg ~ lognormal(mu = -1.60944, sigma = 1)
-          lm_sf_sigma ~ normal(mu = 0, sigma = 1)
+          lm_sf_omega_bsld ~ lognormal(mu = -1.60944, sigma = 1) T[0, ]
```
gravesti, line 32: @gowerc 
For lognormal and gamma does the truncation come automatically? It seems to me it shouldn't do anything in this case.

gravesti (approved):
Thanks @gowerc !

PR #433, opened 2025-02-18
Fixed Example file

### LINKED ISSUE #432: Issue with "scales = 'free'" on longitudinal plots
Is there any way to modify the "facet" argument in autoplot()? 
Note: the below code does not result in free scales unfortunately.
e.g. autoplot(ipred_set, scales="free")+
  labs(x = "Time (years)", y = "SLD (mm)")+
  theme_minimal()

gowerc:
Closes #432 

Note that there is no issue here but I updated the example to use `subject` instead of `pt` so that it actually works again and also added an additional comment to explain how to manipulate the ggplot2 object


gravesti (approved):
Looks good @gowerc

PR #436, opened 2025-02-18
Rename intro vignette

### LINKED ISSUE #435: Rename Vignette
The main introductory vignette is called "model fitting" which I think people often miss. I think this should just be renamed as "introduction to jmpost" to make it more obvious.

gowerc:
Closes #435 

Simple rename + added a introduce to introduce what the vignette is focussed on.


gravesti (dismissed):
Thanks @gowerc 
Could we also set the order of the vignettes in _pkgdown.yaml? I could do that in this PR if you agree

_pkgdown.yml
```diff
@@ -192,3 +192,13 @@ reference:
...
+
+articles:
+  - title: Articles
+    navbar: ~
+    contents:
+        - quickstart
+        - '`statistical-specification`'
+        - '`custom-model`'
+        - '`extending-jmpost`'
+
```
gravesti, line 204: @gowerc Is that a sensible order?

gravesti (approved):

PR #437, opened 2025-02-19
Add saveObject function

### LINKED ISSUE #431: `jmpost` model fit object does not yet contain all sample data
I have hit a problem when uploading model fit objects from `jmpost` to GitHub and then compiling a Quarto doc based on it (while working on https://github.com/RCONIS/tgi-os-training):
It seems not all `cmdstanr` samples data is saved in the object, when just saving the `sampleStanModel` result to disk via a classical `save` function call. I had errors then mentioning missing local file system paths. 
So it would be good to look into the need for using https://mc-stan.org/cmdstanr/reference/fit-method-save_object.html before building the `JointModelSamples` object in https://github.com/Genentech/jmpost/blob/main/R/JointModel.R#L225.

gowerc:
Closes #431 

Added wrapper `saveObject()` method that takes care of ensuring the samples are correctly saving along with the object.

### DIFF
--- a/.github/workflows/check.yaml
+++ b/.github/workflows/check.yaml
@@ -73,9 +73,9 @@ jobs:
       link-checking-method: lychee
       lychee-additional-args: "" # override the default of excluding private links
       lychee-fail: true
-  # version:
-  #   name: Version Check 🏁
-  #   uses: insightsengineering/r.pkg.template/.github/workflows/version.yaml@main
+  version:
+    name: Version Check 🏁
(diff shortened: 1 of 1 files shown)


danielinteractive (dismissed):
Looks great, thanks @gowerc , would it be worth to add a `NEWS` entry?

danielinteractive (dismissed):
Nice! Thanks @gowerc

gravesti (approved):

PR #442, opened 2025-03-13
Add helper function to extract patient level random effects

### LINKED ISSUE #423: Add patient identifiers to the output of the data.frame function (and maybe plots ?)

gowerc:
Closes #423 

I really wanted to be able to just re-use the `Quantities()` group of functions/objects as they are functionally very similar but there was just enough differences in how they are constructed and handled that it become more burden than it was worth to try and marry the code so ended up just splitting up into its own separate object family. Bit verbose but hopefully less complexity as a result. 

Basic Explanation:

- `getRandomEffectsNames()` is a helper method to map the Stan object parameter names to the human readable labels for each model
- `LongitudinalRandomEffects()` is the main user-facing function that users should call on the Joint Model samples object which constructs a `RandomEffectQuantities` object 
- Users can then call `summary()` or `as.data.frame()` on the `RandomEffectQuantities` object to get either the summary stats or the individual samples respectively. 
- Most of the file changes are documentation updates due to the use of `@family` to cross link

EDIT --- 

Looks like the default lintr has been updated to reject explicit returns so there are a few other random edits just to stop it from erroring.

### DIFF
--- a/R/JointModel.R
+++ b/R/JointModel.R
@@ -118,12 +118,10 @@ as.StanModule.JointModel <- function(object, ...) {
     # use the `decorated_render` constants `machine_double_eps`.
     stan_full <- decorated_render(.x = stan_full)
 
-    x <- merge(
+    merge(
         StanModule("base/functions.stan"),
         StanModule(stan_full)
     )
-
-    return(x)
 }
 
 


gravesti (approved):
This looks great @gowerc !

PR #479, opened 2026-08-10
474: compare with fix pars approach

### LINKED ISSUE #474: Compare with approach with fixing parameter values from fitting historical trial data
Please see description from @mercifr1 below.

To do:

- [x] Add `prior_const_vector()` to also allow to give constant values to vectors.
- [x] Add the sequential approach to the TGI-OS training chapter

danielinteractive:
closes #474


danielinteractive:
[1_bhm-jmpost.pdf](https://github.com/user-attachments/files/30934489/1_bhm-jmpost.pdf)

Hi @mercifr1 I tried the sequential approach out in this example, can you double check if this is the idea?...

mercifr1:
Hi @danielinteractive 
As discussed, please consider the following comments:
1) In my understanding, with prior_const(), we set a prior to a point-mass value. Then, we combine it with the likelihood, so the Rhat and ESS_bulk should be estimated, right? I struggle to understand why Rhat and ESS_bulk are NA.
2) In the example, prior_const(_vector) is applied to the TGI model, but we could have applied it to every parameter, right? including the survival parameters and the link function one. Can you confirm?
3) The example contains quite a few data manipulation steps to handle the model outputs. In page 21, it says "prior_const_vector() supplies the arm-specific shrinkage and growth parameters in the same arm order as shown above." The "order" of the arm was actually obtained from a rather complex step involving a call to bind_rows(Map(function(if ... else())). Is there any way to make it more explicit?
THANK YOU!

